PR #10, opened 2023-03-12
cleanup and add tests that run in node and browser

mistermoe:
image


michaelneale:
@mistermoe 13 03 2023 14:52:28.156:INFO [launcher]: Starting browser FirefoxHeadless
13 03 2023 14:52:28.157:ERROR [launcher]: No binary for FirefoxHeadless browser on your platform.

is there an installation guide for that?

is the idea to ensure it works in browser as well as not?

mistermoe:
@michaelneale 

> is the idea to ensure it works in browser as well as not?

yep. `npm run test:node` will run tests in node. `npm run test:browser` will run same tests in chrome, webkit, and firefox.

```
13 03 2023 14:52:28.156:INFO [launcher]: Starting browser FirefoxHeadless
13 03 2023 14:52:28.157:ERROR [launcher]: No binary for FirefoxHeadless browser on your platform.
```
> is there an installation guide for that?

ah good catch. shouldn't have to install firefox to run tests. added playwright's firefox executable path to accommodate. i uninstalled firefox on my machine and ran tests to make sure it worked. can you pull and try again?

michaelneale:
@frankhinek I think this is not a bad starting point - at least lays down some rails. Would like to test a bit more  - but api needs to be stable (also I really don't understand modules in javascript).

frankhinek (approved):
LGTM

Issue #4, opened 2023-03-08
Local storage keys - More unique?


sondreb:
The `connect.js` attempts to read key value "keys", that might be fine when the code always runs on a dedicated site, but it's fairly generic and easy to get conflict (especially for developers running on localhost).

Perhaps have a prefix for the whole library that namespaces all the local storage keys?

frankhinek:
Hey @sondreb -- thanks for highlighting this potential issue.

For anyone reading later, local storage keys are scoped to the specific `protocol://host:port`, so less likely to hit conflicts for a publicly accessible deployment or when using with a desktop Web5 agent, but point taken on the potential issue since devs frequently re-use `localhost:3000`, `localhost:8080`, etc.

frankhinek:
Fixed by #13

PR #11, opened 2023-03-14
Localstorage Keys Prefixed - issue:#4

rpc404:
Defined a constant variable in the library that holds the prefix value and used it throughout the library to prefix all the local storage keys to minimize the chances of conflicts and ensure that the library works reliably across different environments.

### DIFF
--- a/src/connect.js
+++ b/src/connect.js
@@ -1,30 +1,30 @@
+import nacl from "tweetnacl";
+import * as SDK from "@tbd54566975/dwn-sdk-js";
 
-import nacl from 'tweetnacl';
-import * as SDK from '@tbd54566975/dwn-sdk-js';
+const localStorageKeyPrefix = "web5_";
 
 /* Keys */
 
 function getKeys() {
-  let keys = JSON.parse(localStorage.getItem('keys') || null);
+  let keys = JSON.parse(localStorage.getItem(localStorageKeyPrefix + "keys") || null);
   if (keys) {
     keys = {
       encoded: keys,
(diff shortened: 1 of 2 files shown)


src/connect.js
```diff
@@ -1,38 +1,38 @@
+import nacl from "tweetnacl";
```
frankhinek, line 1:
```suggestion
import nacl from 'tweetnacl';
```

frankhinek:
Hey @rpc404 -- thanks for submitting this PR to address the potential for local storage key conflicts.

A few asks before I can complete the review and merge:
- Remove formatting changes from README
- Switch all double-quotes back to single-quote -- convention for all of our libraries and you'll see the rule in the eslint file
- Remove trailing commas for last item in array/object
- Remove indentation changes for the switch statements
- Remove changes to line wrapping

Thanks

rpc404:
Suggestion updated.

michaelneale:
LGTM

frankhinek:
Hey @rpc404 -- sincerely appreciate your contribution and follow-up.  This library is still in a pre-beta stage and there was a major refactoring that has been pending review & approval for the past week+.  With the merging of PR #13, a modular storage approach was introduced which both uses a "web5-" prefixed browser storage key and also enables the developer to use a custom storage option of their choosing.

As a result of the fairly significant refactor, the key collision issue was resolved.

The API should be stable going forward and it will be very unlikely for such a significant refactor to occur, so we'd welcome your questions on Discord and additional code contribution.

PR #15, opened 2023-03-16
an example thanks to moe and mic

michaelneale:


frankhinek:
Thanks, Mic, for prodding to get in-browser working. 

PR #16 fixed issues that were preventing in-browser DWNs from working.

Going to add examples using Web5-js in https://github.com/TBD54566975/incubating-web5-labs

PR #22, opened 2023-03-29
Add Simple Agent example

frankhinek:
This PR is purely a documentation / examples change:
- Move `test-site` into `examples/` and rename to `test-dashboard`
- Add a `simple-agent` example with a README


csuwildcat (approved):

michaelneale:
@frankhinek is the term "agent" to be used instead of server now?

csuwildcat:
I would still distinguish the two,, because an agent has keys and isn't necessarily an HTTP server, while a server is and doesn't have the user's keys.

michaelneale:
ah - thanks @csuwildcat that makes sense

PR #24, opened 2023-04-03
add support for sending permission requests

dcrousso:


dcrousso:
handled by #37 and #40

PR #26, opened 2023-04-05
adding a release workflow - similar to dwn sdk

michaelneale:
cc @frankhinek @mistermoe - this is what DWN used which should work similarly for web5-js


frankhinek (approved):
LGTM.  Thanks @michaelneale

Issue #31, opened 2023-04-06
Create convenience method for web5.did.create('ion')


ALRubinger:
Nothing broken here, but an opportunity for Web5 to erase complexity and be more approachable to newcomers. 

One of the first things we do in the QuickStart is have the user create a DID. When using the recommended:

```
let did = await web5.did.create('ion');
```

...we are left to immediately delve into conceptual topics that we think would make a stronger experience if we could avoid linking into those concepts:

* Why we pass a `String` in there
* What `ion` is
* What a DID Method Name is 

Is there some syntactic sugar we can add to the Web5 API which erases this? Perhaps:

* `web5.did.create();` // No-arg constructor which applies a default we choose, like `ion`?
* `web5.did.createIon();` // An Autocomplete version that tooling can help with so we don't need to pass a manual String?
* `web5.did.create(Web5.DID_METHODS.ION);` // A constructor that accepts a type and an enum of those types

Open to other ideas - whatever makes for easier autocomplete and intuitive API, and limits the conceptual weight of someone having to make a choice about DID method types, understanding why, and passing in an untyped String to designate which.

frankhinek:
Hey @ALRubinger -- good question and callout.

My preference is:
- `web5.did.create();` // No-arg constructor which applies a default we choose, like ion?

I'd be ok with making the default `ion` and would be simple to implement.

@csuwildcat you cool with that?

csuwildcat:
I personally would prefer not to have a default, in case we ever swap that default. This is a really impactful thing to default, and this would mean a significant breaking change for all downstream lib users if we ever did it.

dcrousso:
i think at the very least we should at least have something like `web5.did.MethodName.Ion` so that developers don't have to know any magic strings (as well as having an easy way for the developer to log in the `console` what all the method names are)

having a default would be nice as it could let developers just DoTheThing™ `web5.did.create()` and not even have to worry about a method.

@csuwildcat i do get your concern about us being constrained in changing the default.  what if we had a `web5.did.DefaultMethodName` so that developers who do care about the default can know what it is and do logic based on it without having do awful stuff like looking at the source or checking the package version

dcrousso:
i went ahead with the first part at least :) https://github.com/TBD54566975/web5-js/pull/32

mistermoe:
[...]

PR #32, opened 2023-04-07
create/use enums instead of magic strings

dcrousso:


mistermoe:
@ALRubinger we can get most of the benefits that we'd get from enums just by using `JSDoc`. almost all code editors read JSDoc and use it to provide intellisense. e.g. 

image

dcrousso:
i still think there's some benefit in the developer being able to check at runtime what things they could provide to `web5.did.create()` without having to call it and check if it throws

ALRubinger:
Thanks! This is awesome for removing the need for free-floating Strings. It does still force the user to a decision about something they may not understand or care about: choosing a DID method. Made my case on opting for a sensible default there in #31 - thanks y'all!

ALRubinger:
We're shipping the Quickstart without this in place; when done, please open an issue on the developer.tbd.website repo to update the Quickstart w/ this syntax - thanks!

dcrousso:
handled by #40

PR #33, opened 2023-04-07
make `key` the default DID method when calling `web5.did.create()`

dcrousso:
this will allow developers who just want to get a basic DID to do so without having to even know anything about what a DID method is

it still lets them "do better" by providing a DID method to get more features (e.g. endpoints)

### DIFF
--- a/src/did/Web5DID.js
+++ b/src/did/Web5DID.js
@@ -15,7 +15,7 @@ class Web5DID {
     return this.#web5;
   }
 
-  async create(method, options = { }) {
+  async create(method = 'key', options = { }) {
     const api = await this.#getMethodAPI(method);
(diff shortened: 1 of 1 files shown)


src/did/Web5DID.js
```diff
@@ -15,7 +15,7 @@ class Web5DID {
     return this.#web5;
   }
 
-  async create(method, options = { }) {
+  async create(method = 'key', options = { }) {
```
mistermoe, line 18: can we go ahead and add JSDoc here? re: https://github.com/TBD54566975/web5-js/pull/32#issuecomment-1500401159

dcrousso:
handled by #40

PR #36, opened 2023-04-13
fix dist paths

mistermoe:


mistermoe:
@frankhinek

frankhinek (approved):
Thanks for catching and fixing @mistermoe

PR #40, opened 2023-04-15
cleanup and fixes after #37

dcrousso:


examples/simple-agent/src/utils.js
```diff
@@ -51,7 +51,7 @@ async function loadConfig() {
 }
 
 async function initOperatorDid() {
-  const operatorDid = await web5.did.create('ion', {
+  const operatorDid = await web5.did.create(web5.did.DIDMethodName.ION, {
```
michaelneale, line 54: not sure this is better - as did methods are "publicly known", and closer to domain names in a way than types (they are types but there are already quite a lot of them).

dcrousso: interesting.  didnt know that

i think this structure offers a few benefits

- allows for callers to know what methods are supported by the version of `web5-js` they import by looking at whether a key exists in `web5.did.DIDMethodName`
- enables minifiers/bundlers to avoid including raw strings more than once
- doesn't preclude callers from providing a string, as ultimately that's what `web5.did.DIDMethodName.*` resolves to

dcrousso: also when you say "closer to domain names" does that mean we should be doing some sort of normalization on the inputs (i.e. `method = method.toLowerCase()`)?

i ask cause i wouldnt necessarily know whether to use `"ion"` or `"ION"` since the latter is what the source website uses, whereas the former is what we use :/

michaelneale: yes I think should be normalised probably.

Issue #35, opened 2023-04-12
case issue: keypair vs keyPair


jdutchak:
in web5.js the key pair param is lower case "keypair"

in dwn-sdk-js the param is keyPair - camelCase - when trying to createSignatureInput or ProtocolsConfigure.create this becomes an issue and it appears the key method doesn't include like the "alg" param like the ion method.

suggest ensuring case is same from dwn-sdk-js or ssi -> other projects (preferred camelCase in this scenario)

PR #43, opened 2023-04-19
Add Record class, DID Manager, and Stabilize API

### LINKED ISSUE #20: Ensure all dwn.METHOD() APIs return transport agnostic results and data
Some of our APIs currently leak the HTTP response wrapper out to the developer, and that means a dev would have to handle whatever comes out from those methods in a transport-specific way. Instead we should normalize all the outputs so they are identical regardless of whatever transport the internal calls were made over.

### LINKED ISSUE #27: Function name, did.register, is confusing
Registering a DID has a well known meaning. 

In the context of web5-js, `web5.did.register` has a different meaning - it's associating a DID with a DWN.

This will lead to a confusing developer experience. Please change the name of this function to a term that's not overloaded

frankhinek:
### External API changes
- replace `web5.did.register()` with `web5.did.manager()`
    - https://github.com/TBD54566975/web5-js/blob/4362ce22d5ea77c280c1e6eb17726b6a37af51de/src/did/manager.js
    - https://github.com/TBD54566975/web5-js/blob/4362ce22d5ea77c280c1e6eb17726b6a37af51de/src/did/web5-did.js#L41
    - Fixed #27 
- New `Record` class which greatly improves the developer experience and makes working with records and record data significantly easier.
    - https://github.com/TBD54566975/web5-js/blob/4362ce22d5ea77c280c1e6eb17726b6a37af51de/src/dwn/models/record.js
- Add `web5.dwn.records.createFrom()` as a convenience to make it easy to create a new record based on an existing record.
    - https://github.com/TBD54566975/web5-js/blob/4362ce22d5ea77c280c1e6eb17726b6a37af51de/src/dwn/interfaces/records.js#L16
- Add `web5.dwn.records.create()` which is an alias for `web5.dwn.records.write()`
    - https://github.com/TBD54566975/web5-js/blob/4362ce22d5ea77c280c1e6eb17726b6a37af51de/src/dwn/interfaces/records.js#L12

### Internal function changes:
- `#createSignedMessage()` now takes an `author` parameter which removes redundant data from being stored by the DidManager and a pre-req for future improvements to key management.
https://github.com/TBD54566975/web5-js/blob/4362ce22d5ea77c280c1e6eb17726b6a37af51de/src/Web5.js#L80
- Replace `delete` methods on `Storage` class and inherited classes with `remove` to make it possible to create functional storage adapters (`delete` is a reserved word if used outside of class methods).

### Chores/Cleanup
- Use `dwn` key ID by default for ION dids created with `web5.did.create()` https://github.com/TBD54566975/web5-js/blob/4362ce22d5ea77c280c1e6eb17726b6a37af51de/src/did/methods/ion.js#L9
- Resolve `keypair` / `keyPair` discrepancy raised in #35
- Standardize all files, variables, and classes to kebab, camel, and Pascal case, respectively to align with naming convention in `dwn-sdk-js`
- Align use of HTTP 400 series status codes with `dwn-sdk-js`
- Fixes #20 

Note to @TBD54566975/tbd-developer-relations:
- No doc work for changes to the `examples/` directory, which are currently used for validating/testing during development


src/did/connect/connect.js
```diff
@@ -77,53 +76,52 @@ export class DIDConnect {
...
+      case DidConnectStep.Delegation: {
         const delegationResult = rpcMessage?.result;
 
         // Success
         if (delegationResult?.ok) {
           const authorizedDid = delegationResult?.message?.grantedBy;
-          // Register DID now that the connection was authorized
-          await this.web5.did.register({
+          // Set Managed DID now that the connection was authorized
+          await this.web5.did.manager.set(authorizedDid, {
```
michaelneale, line 123: so a managed did is the one that is the "main" one for the DWN? do I have that right?

src/dwn/web5-dwn.js
```diff
@@ -53,5 +53,5 @@ class Web5DWN {
 }
 
 export {
-  Web5DWN,
+  Web5Dwn,
```
michaelneale, line 56: similar here - this is how we CamelCase?

examples/test-dashboard/simple-agent.html
```diff
@@ -112,6 +112,7 @@ <h2>Read Records</h2>
 
   <script type="module">
     import { Web5 } from './browser.mjs';
+    import { DataStream } from 'https://cdn.jsdelivr.net/npm/@tbd54566975/dwn-sdk-js@0.0.26';
```
michaelneale, line 115: @frankhinek @mistermoe this will be an extra load from the CDN, could that datastream be bundled into web5 js?

michaelneale (approved):
Looks nice. 

Main outstanding question was additional pull of the DWN (with its own version) to work in a browser which seemed a bit odd (see comment on the simple-agent.html)

src/did/web5-did.js
```diff
@@ -37,8 +38,18 @@ class Web5DID {
     return this.#didConnect.permissionsRequest;
   }
 
+  get manager() {
```
dcrousso, line 41: this means that every time someone calls `web5.did.manager` they're gonna get a new object.  can we avoid this?  it seems very unnecessary

also, `DidManager` seems kinda pointless since all it really does is just call methods on the `store`.  do we really need that collection of functions instead of just `this.didStore.thing(arg)`?

src/did/web5-did.js
```diff
@@ -1,19 +1,20 @@
...
-class Web5DID {
+export class Web5Did {
   #cryptoCiphers = {};
   #didConnect;
   #web5;
 
-  #registeredDIDs = new MemoryStorage();
-  #resolvedDIDs = new MemoryStorage();
+  #resolvedDids = new MemoryStorage();
+  didStore = new MemoryStorage();
```
dcrousso, line 17: should this be private or public?

src/did/connect/connect.js
```diff
@@ -1,17 +1,16 @@
-import { DIDConnectRPCMethods, DIDConnectStep, JSONRPCErrorCodes, findWebSocketListener } from './utils.js';
+import { DidConnectRpcMethods, DidConnectStep, JsonRpcErrorCodes, findWebSocketListener } from './utils.js';
 import { WebSocketClient } from './ws-client.js';
-import { parseJSON } from '../../utils.js';
-import { LocalStorage } from '../../storage/LocalStorage.js';
+import { parseJson } from '../../utils.js';
+import { LocalStorage } from '../../storage/local-storage.js';
 
-export class DIDConnect {
+export class DidConnect {
```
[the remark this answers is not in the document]
dcrousso: yeah why is this PR making all these style changes?

src/dwn/dwn-utils.js
```diff
@@ -0,0 +1,7 @@
+/**
+ * Uses duck typing to determine whether the stream is a web browser ReadableStream
+ * or a Node.js Readable stream.
+ */
+export function isReadableWebStream(stream) {
+  return typeof stream._read !== 'function';
```
dcrousso, line 6: can we not do something like `globalThis.ReadableStream && stream instanceof globalThis.ReadableStream` instead?

src/dwn/models/record.js
```diff
@@ -0,0 +1,244 @@
...
+  #isFrozen = false;
+  #readableStream = null;
+
+  constructor(dwn, options = { }) {
+    this.#dwn = dwn;
+
+    // RecordsWriteMessage properties.
+    const { author, contextId = undefined, descriptor, recordId = null, target } = options;
+    this.#contextId = contextId;
+    if (descriptor?.data) delete descriptor.data;
```
dcrousso, line 25: can we do `delete this.#descriptor.data;` below instead?  then we dont have to check for anything

src/dwn/models/record.js
```diff
@@ -0,0 +1,244 @@
...
+   */
+
+  get data() {
+    if (!this.#encodedData && !this.#readableStream) {
+      // `encodedData` will be set if `dataSize` <= DwnConstant.maxDataSizeAllowedToBeEncoded. (10KB as of April 2023)
+      // `readableStream` will be set if Record was instantiated from a RecordsRead reply.
+      // If neither of the above are true, then the record must be fetched from the DWN.
+      this.#readableStream = this.#dwn.records.read(this.#target, { author: this.#author, message: { recordId: this.#recordId } })
+        .then((response) => response.record )
+        .then((record) => { return record.data; });
```
dcrousso, line 81: why not just `(record) => record.data`?

src/dwn/models/record.js
```diff
@@ -0,0 +1,244 @@
...
+      // type is Base64 URL encoded string if the Record object was instantiated from a RecordsQuery response
+      // If it is a string, we need to Base64 URL decode to bytes
+      this.#encodedData = Encoder.base64UrlToBytes(this.#encodedData);
+    }
+
+    const self = this; // Capture the context of the `Record` instance.
+    const dataObj = {
+      async json() {
+        if (self.#encodedData) return this.text().then(JSON.parse);
+        if (self.#readableStream) return this.text().then(JSON.parse);
```
dcrousso, line 97: it's really odd that we use `self` and then `this`.  can we rework this to make it less confusing?  ditto below

also, can we combine these into a single `if`?

src/dwn/models/record.js
```diff
@@ -0,0 +1,244 @@
...
+        if (self.#encodedData) return Encoder.bytesToString(self.#encodedData);
+        if (self.#readableStream) return self.#readableStream.then(DataStream.toBytes).then(Encoder.bytesToString);
+        return null;
+      },
+      async stream() {
+        if (self.#encodedData) return DataStream.fromBytes(self.#encodedData);
+        if (self.#readableStream) return self.#readableStream;
+        return null;
+      },
+      then (...callbacks) {
```
dcrousso, line 110: Style: no need for a space before the `(`

src/dwn/models/record.js
```diff
@@ -0,0 +1,244 @@
...
+      // If `encodedData` is set, then it is expected that:
+      // `dataSize` <= DwnConstant.maxDataSizeAllowedToBeEncoded (10KB as of April 2023)
+      // type is Uint8Array bytes if the Record object was instantiated from a RecordsWrite response
+      // type is Base64 URL encoded string if the Record object was instantiated from a RecordsQuery response
+      // If it is a string, we need to Base64 URL decode to bytes
+      this.#encodedData = Encoder.base64UrlToBytes(this.#encodedData);
+    }
+
+    const self = this; // Capture the context of the `Record` instance.
+    const dataObj = {
```
dcrousso, line 94: this means we create a new object anytime someone calls `.data`.  can we avoid this?

src/dwn/web5-dwn.js
```diff
@@ -1,13 +1,13 @@
...
-import { Permissions } from './interface/Permissions.js';
-import { Protocols } from './interface/Protocols.js';
-import { Records } from './interface/Records.js';
+import { Permissions } from './interfaces/permissions.js';
+import { Protocols } from './interfaces/protocols.js';
+import { Records } from './interfaces/records.js';
 import { createWeakSingletonAccessor } from '../utils.js';
 
-const sharedNode = createWeakSingletonAccessor(() => SDK.Dwn.create());
+const sharedNode = createWeakSingletonAccessor(() => Sdk.Dwn.create());
```
dcrousso, line 8: I dont think ive ever seen "SDK" written as "Sdk" :/

src/dwn/web5-dwn.js
dcrousso: yeah i really dislike this

src/storage/local-storage.js
```diff
@@ -9,7 +9,7 @@ class LocalStorage extends Storage {
     localStorage.setItem(key, JSON.stringify(value));
   }
 
-  async delete(key) {
+  async remove(key) {
```
dcrousso, line 12: why was this renamed?

src/did/web5-did.js
```diff
@@ -174,13 +172,4 @@ class Web5DID {
...
-
-  /**
-   * Utility functions for working with DIDs
-   */
-  #util = { ...DidUtils };
 }
-
-export {
-  Web5DID,
-};
```
dcrousso, line 186: why was this changed?

src/transport/http-transport.js
```diff
@@ -41,17 +42,21 @@ class HTTPTransport extends Transport {
...
+
+    if (!response.ok) {
+      throw new Error(`Fetch failed with status ${response.status}`);
+    }
+
+    const web5ResponseHeader = response.headers.get(this.ENCODED_RESPONSE_HEADER);
+    if (web5ResponseHeader) {
+      // RecordsRead responses return `message` and `status` as header values, with a `data` ReadableStream in the body.
+      const { entries = null, message, record, status } = await this.decodeMessage(web5ResponseHeader);
+      return { entries, message, record: { data: response.body, ...record }, status };
```
dcrousso, line 55: this means that if `record` has a `data` property it will override the `data` you specify.  id suggest putting the `...record` first so that the `data: response.body` will never be overriden

src/dwn/interfaces/records.js
```diff
@@ -0,0 +1,103 @@
...
+  constructor(dwn) {
+    super(dwn, 'Records');
+  }
+
+  get create() {
+    return this.write;
+  }
+
+  async createFrom(target, request) {
+    const { author: inheritedAuthor, target: _, ...inheritedProperties } = request.record.toJSON();
```
dcrousso, line 17: this seems like a super subtle way of not including `target` in `inheritedProperties`

can we do an `delete inheritedProperties.target;` somewhere down below instead?  that seems way more explicit (and gives an opportunity to comment as to why)

src/dwn/interfaces/records.js
```diff
@@ -0,0 +1,103 @@
...
+import { dataToBytes } from '../../utils.js';
+
+export class Records extends Interface {
+  constructor(dwn) {
+    super(dwn, 'Records');
+  }
+
+  get create() {
+    return this.write;
+  }
```
dcrousso, line 14: i dont think this will propagate `this` correctly.  id suggest having it be this instead
```
async create(...args) {
  return this.write(...args);
}
```
that will also make it appear correctly in the Web Inspector (i.e. a function instead of a getter)

src/dwn/interfaces/records.js
```diff
@@ -0,0 +1,103 @@
...
+
+  async createFrom(target, request) {
+    const { author: inheritedAuthor, target: _, ...inheritedProperties } = request.record.toJSON();
+
+    // If `data` is being updated then `dataCid` and `dataSize` must not be present.
+    if (request?.data !== undefined) {
```
dcrousso, line 20: there's no need to do `request?.` as you've already assumed it's an object above by `request.record`

ditto below

src/dwn/interfaces/records.js
```diff
@@ -0,0 +1,103 @@
...
+    // If `published` is set to false, ensure that `datePublished` is undefined. Otherwise, DWN SDK's schema validation
+    // will throw an error if `published` is false but `datePublished` is set.
+    if (request?.message?.published === false && inheritedProperties?.datePublished !== undefined) {
+      delete inheritedProperties.datePublished;
+      delete inheritedProperties.published;
+    }
+
+    // If the request changes the `author` or message `descriptor` then the deterministic `recordId` will change.
+    // As a result, we will discard the `recordId` if either of these changes occur.
+    if ((request?.message && Object.keys(request.message).length > 0)
```
dcrousso, line 34: can we use `!isEmptyObject(request.message)` instead?

src/dwn/interfaces/records.js
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    return { ...response, record };
+  }
+
+  async query(target, request) {
+    const response = await this.send('Query', target, request);
+
+    const entries = [];
+    response.entries.forEach(entry => {
```
dcrousso, line 69: why not use `Array.prototype.map`?
```
const entries = response.entries.map(entry => new Record(this.dwn, { ...entry, target, author: request.author }));
```

src/dwn/models/record.js
```diff
@@ -0,0 +1,244 @@
...
+    // If the record was created from a RecordsRead reply then it will have a `data` property.
+    if (options?.data) {
+      // this.#readableStream = isReadableWebStream(options.data) ? toIsomorphicReadableStream(options.data) : options.data;
+      this.#readableStream = isReadableWebStream(options.data) ? new ReadableWebToNodeStream(options.data) : options.data;
+    }
+  }
+
+  // Mutable Web5 Record Class properties.
+  get author() { return this.#author; }
+  get isFrozen() { return this.#isFrozen; }
```
dcrousso, line 46: you say that this is mutable but i dont see a `set isFrozen`?  is that actually mutable then?

src/dwn/models/record.js
```diff
@@ -0,0 +1,244 @@
...
+
+  async delete() {
+    if (this.isFrozen) throw new Error(`Error: Record with ID '${this.id}' was previously deleted.`);
+
+    // Attempt to delete the record from the DWN.
+    const response = await this.#dwn.records.delete(this.#target, { author: this.#author, message: { recordId: this.#recordId } });
+    
+    if (response.status.code === 202) {
+      // If the record was successfully deleted, freeze the instance to prevent further modifications.
+      this.#freezeRecord();
```
dcrousso, line 132: i would suggest `#markAsDeleted` (and `#isDeleted`) as that's more clear as to what it does

src/dwn/models/record.js
```diff
@@ -0,0 +1,244 @@
...
+      datePublished: this.datePublished,
+      published: this.published,
+    };
+  }
+
+  #freezeRecord() {
+    this.#isFrozen = true;
+  }
+
+  #unFreezeRecord() {
```
dcrousso, line 232: this doesn't appear to be used.  do we actually want to allow for this?

src/dwn/models/record.js
```diff
@@ -0,0 +1,244 @@
...
+    this.#isFrozen = false;
+  }
+
+  static #verifyPermittedMutation(propertiesToMutate, mutableDescriptorProperties) {
+    propertiesToMutate.forEach(propertyName => {
```
dcrousso, line 237: id use `for (const property of propertiesToMutate)` as that would allow for this to be called with an array or `Set` or etc. (i.e. any flat iterable) instead of just an array

src/dwn/models/record.js
```diff
@@ -0,0 +1,244 @@
...
+      if (!mutableDescriptorProperties.includes(propertyName)) {
+        throw new Error(`${propertyName} is an immutable property. Its value cannot be changed.`);
+      }
+    });
+    return true;
```
dcrousso, line 242: does this actually need to return anything?  it seems like it's just "success if it doesnt `throw`"

src/dwn/models/record.js
```diff
@@ -0,0 +1,244 @@
...
+  async update(options = { }) {
+    if (this.isFrozen) throw new Error(`Error: Record with ID '${this.id}' was previously deleted.`);
+
+    // Begin assembling update message.
+    let updateMessage = { ...this.#descriptor, ...options };
+
+    // If `data` is being updated then `dataCid` and `dataSize` must be undefined and the `data` property is passed as
+    // a top-level property to `web5.dwn.records.write()`.
+    let data;
+    if (options?.data !== undefined) {
```
dcrousso, line 147: are we allowing for a scenario where `options` doesnt exist even though we've given it a default value?

AFAIK the only way for that to be possible is if the caller explicitly provides `null` or a non-object value (in which case we probably should be checking for that, no?)

examples/test-dashboard/simple-agent.html
dcrousso: we also could just load from the JS in `dist` instead

src/web5.js
```diff
@@ -71,18 +71,19 @@ class Web5 extends EventTarget {
...
-    return { status: { code: 422, detail: 'Target DID could not be resolved' } };
+    return { status: { code: 400, detail: 'Target DID could not be resolved' } };
   }
 
-  async #createSignedMessage(resolvedAuthor, message, data) {
-    const authorizationSignatureInput = this.#dwn.SDK.Jws.createSignatureInput({
-      keyId: resolvedAuthor.did + '#key-1',
-      keyPair: resolvedAuthor.keys,
+  async #createSignedMessage(author, resolvedAuthor, message, data) {
+    const keyId = '#dwn';
```
dcrousso, line 81: this seems like a really odd choice of key name.  is the idea to try to hide this from callers?  if so, can we use a `Symbol` instead?

src/dwn/models/record.js
```diff
@@ -0,0 +1,244 @@
...
+   */
+  toJSON() {
+    return {
+      author: this.author,
+      target: this.target,
+      recordId: this.id,
+      contextId: this.contextId,
+      dataFormat: this.dataFormat,
+      dateCreated: this.dateCreated,
+      interface: this.interface,
```
dcrousso, line 214: are we trying to emulate the JSON structure of what's in `dwn-sdk-js`?  if so, shouldn't some of these be in a `descriptor` (sub)object?

src/types.js
```diff
@@ -0,0 +1,230 @@
...
+ * NOTE: These are temporary until Web5 JS can be converted to TypeScript and import the types from dwn-sdk-js.
+ */
+
+/**
+ * Web5 JS type definitions.
+ */
+
+/**
+ * @typedef {Object} Web5SendResponseMessage
+ * @property {ProtocolsConfigureDescriptor | ProtocolsQueryDescriptor | RecordsQueryDescriptor | RecordsReadDescriptor | RecordsWriteDescriptor} message
```
dcrousso, line 11: a bit unfortunate that we have all these other types in a separate file instead of right next to where the actual types are declared/used

id hate to have to remember to also update this file whenever im changing something elsewhere in the repo :/

src/web5.js
```diff
@@ -71,18 +71,19 @@ class Web5 extends EventTarget {
       if (dwnNodes) {
         return this.#send(dwnNodes, { author, data, message, target });
       }
-      return { status: { code: 422, detail: 'No DWN endpoints present in DID document. Request cannot be sent.' } };
+      return { status: { code: 400, detail: 'No DWN endpoints present in DID document. Request cannot be sent.' } };
```
dcrousso, line 74: why did this change?  "Unprocessable Content" seems like a reasonable response for this, no?

src/web5.js
```diff
@@ -108,25 +109,29 @@ class Web5 extends EventTarget {
...
-    return response ?? { status: { code: 503, detail: 'Service Unavailable' } };
+    if (!isUnsignedMessage(request.message)) {
+      // If the message is signed return the `descriptor`, and if present, `recordId`.
+      const { recordId = null, descriptor } = request.message.message;
+      message = { recordId, descriptor };
+    }
+
+    response ??= { status: { code: 503, detail: 'Service Unavailable' } };
+
+    return { message, ...response };
```
dcrousso, line 135: do we want to allow for `message` to be overridden by a `response.message`?

dcrousso (changes requested):

examples/test-dashboard/simple-agent.html
michaelneale: no I am thinking more when loading web5 from the CDN (not for this specific example, as loading from dist won't help?)

examples/simple-agent/src/utils.js
```diff
@@ -1,10 +1,22 @@
...
 
-const web5 = new Web5();
+// Use custom names for the block, message, and data stores to make it possible to launch multiple simple agents
+// in the same directory.  If you don't do this, you will get LevelDB lock errors.
+const port = await getPort(process.argv);
+const dataStore = new DataStoreLevel({ blockstoreLocation: `DATASTORE-${port}` });
+const messageStore = new MessageStoreLevel({
+  blockstoreLocation : `MESSAGESTORE-${port}`,
+  indexLocation      : `INDEX-${port}`,
+});
```
mistermoe, line 15: i believe we'll need to do the same for `eventlog` and `index`

src/did/web5-did.js
michaelneale: `// Set Managed DID as running in browser memory and for which keys are available to sign messages.
    web5.did.manager.set(myDid.id, {`

that looks pretty nice and explanatory to me - keeping things namespaced.

michaelneale:
The style changes seem consistent with https://google.github.io/styleguide/jsguide.html#naming-camel-case-defined - which is more rigorous thing I can find out there (took me a while to dig it up from memory). 

image

package.json
```diff
@@ -69,9 +69,10 @@
   },
   "dependencies": {
     "@decentralized-identity/ion-tools": "1.0.6",
-    "@tbd54566975/dwn-sdk-js": "0.0.26",
+    "@tbd54566975/dwn-sdk-js": "0.0.30-unstable-2023-04-15-149c713",
```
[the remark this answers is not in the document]
mistermoe: @frankhinek i just published `0.0.30`. should be good to go on your end if you want to remove `-unstable-2023-04-15-149c713`

examples/test-dashboard/simple-agent.html
mistermoe: @michaelneale i believe it already is, i think it's `web5.dwn.sdk.DataStream` but i could be wrong

src/did/connect/utils.js
```diff
@@ -103,4 +103,12 @@ export async function findWebSocketListener(
       }
     }
   }
-}
\ No newline at end of file
+}
+
+export async function triggerProtocolHandler(url) {
```
mistermoe, line 108: hoping to take a stab at adding `JSDoc` to this before you merge but don't let this hold you back

examples/test-dashboard/simple-agent.html
frankhinek: This was purely for testing and can be ignored.  I need to do some work to update `simple-agent`

package.json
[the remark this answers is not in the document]
frankhinek: You can ignore this in `simple-agent` Mic -- it was only for testing.  Thanks @mistermoe -- will update to `0.0.30`

src/did/connect/connect.js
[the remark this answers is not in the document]
frankhinek: Per the PR Note:
> Standardize all files, variables, and classes to kebab, camel, and Pascal case, respectively to align with naming convention in dwn-sdk-js

@dcrousso  DevRel team needs to make documentation changes and all of the methods need to have JS Doc. Didn't want to have to do this twice.  The inconsistent use of casing throughout Web5 JS was a mess and this change brings it inline with the underlying `dwn-sdk-js` that it relies on.

src/did/connect/connect.js
frankhinek: A "managed" DID is one that the application/agent has been control of either directly (has access to keys or a signing function) or indirectly through an a remote application/agent.

All other DIDs are assumed to be controlled by other entities and are resolved in the normal manner.  Resolution varies depending on the DID method (`did:key`, `did:web`, `did:ion`).

src/dwn/web5-dwn.js
frankhinek: This is exactly how camel casing works in `dwn-sdk-js`

The actual class in `dwn-sdk-js` is literally `Dwn`:
https://github.com/TBD54566975/dwn-sdk-js/blob/07bcddb399d908fd431f4769dd0e48f1f03d9e68/src/dwn.ts#L28

It was completely inconsistent through Web5 JS previously.  It is now consistent and both libraries match the approach to casing.

src/storage/local-storage.js
frankhinek: `delete` is a reserved word outside of classes and there are non-class based libraries we may end up using for other functionality within the library that won't be able to use delete

src/did/web5-did.js
frankhinek: Use of inline exports and separate exports was inconsistent throughout Web5 JS. Adopted the style used throughout `dwn-sdk-js`.  More work needed to update the rest of Web5 JS but this is purely a style difference.

src/did/connect/utils.js
frankhinek: Definitely in need of lots of JSDoc attention.  There were 0 comments before so trying to add as we go along.  That function didn't have any comments previously either... was moved from the general `utils.js` since it is specific only to DID Connect and I want to make it possible to have DID Connect be an optional import in the future.

src/web5.js
frankhinek: Agreed.  Reverted back to `422`

src/dwn/models/record.js
frankhinek: You are correct that it doesn't need to return anything.

In an earlier iteration, I was re-using `verifyEqualityOfImmutableProperties()` from `dwn-sdk-js`:
https://github.com/TBD54566975/dwn-sdk-js/blob/07bcddb399d908fd431f4769dd0e48f1f03d9e68/src/interfaces/records/messages/records-write.ts#L553-L575

but ended up not wanting to have the same checks.  The `return true` stuck around after the refactor since it is in `verifyEqualityOfImmutableProperties()`.

We can remove the return but should likely update in both `web5-js` and `dwn-sdk-js` if we do.

Thoughts @dcrousso ?

src/web5.js
frankhinek: FWIW, the key name was statically defined as `key-1` before, so I'd argue `#dwn` is "less odd" than `key-1` 😄 

Two things to keep in mind:
1. We haven't introduced a key store to Web5 JS and or at least made it possible for a developer to provide their own.  The key store work has a number of open questions we need to work through, and that will require refactoring all of the methods that touch keys.  We could opt to address then.
2. Since the key store work is still pending, we have been relying on the `keyId` specified when the ION DID is created to match the `keyId` in the `createSignedMessage()` method.  If there's a mismatch, any calls that require message signing fail.  To really make an improvement here what we ought to do is select from one of:
    - Make the `keyId` used for signing messages configurable on a per DID basis WITHOUT requiring it to be passed in on every call to a `web5.dwn.records.*`, `web5.dwn.protocols.*`, etc. call.
    - Make the `keyId` a "well known" that could be a Symbol which is used consistently in both places.  This would mean that an existing DID could not be imported unless it conforms to this approach.  There may be some consideration for doing this even though its restrictive based on how encryption works in `dwn-sdk-js`. 

There is some discussion needed here with @csuwildcat

src/dwn/models/record.js
frankhinek: While the ability to call `JSON.stringify(myRecord)` is an added bonus and the reason why the method is `toJSON()` and not `toJson()` ...

The real reason for adding this was to be able to use with the other `web5.dwn.records.*` methods.  It is used specifically in `createFrom()`:

https://github.com/TBD54566975/web5-js/blob/4362ce22d5ea77c280c1e6eb17726b6a37af51de/src/dwn/interfaces/records.js#L16-L47

To the root of your question as to why have all of the properties be top-level and not nested under `descriptor` the real reason is that `RecordsWrite.create()` takes an `options` prop bag that has all props as top-level

https://github.com/TBD54566975/dwn-sdk-js/blob/07bcddb399d908fd431f4769dd0e48f1f03d9e68/src/interfaces/records/messages/records-write.ts#L139-L225

And consequently, so does the `message` portion of every `web5.dwn.records.*` method.

frankhinek:
> The style changes seem consistent with https://google.github.io/styleguide/jsguide.html#naming-camel-case-defined - which is more rigorous thing I can find out there (took me a while to dig it up from memory).
> 
> image

@michaelneale FWIW, this also isn't introducing anything new that wasn't already the case in `dwn-sdk-js`.  It is now consistent casing between the two libraries.

package.json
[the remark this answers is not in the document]
frankhinek: Although it shouldn't have presented any issues, I tested again with `0.0.30` because it did include one additional commit  beyond `0.0.30-unstable-2023-04-15-149c713`.  No issues, as expected, so bumped to 0.0.30.

mistermoe (approved):
🎖️

src/web5.js
frankhinek: Opened Issue #44 to track this and make a decision.

src/did/web5-did.js
frankhinek: it can be private.  Addressed in https://github.com/TBD54566975/web5-js/pull/43/commits/a01471f36de09561c63155887265985b717fc000

src/dwn/models/record.js
frankhinek: Yes. This was done intentionally because it is possible to undelete a record in a DWN by processing another RecordsWrite with the exact same properties and the tombstoned record.

src/types.js
frankhinek: Fair request although I kept the types together because they'll be gone once the lib is converted to TS and most of these types can be imported from `dwn-sdk-js`.  For now, I moved the two that are Web5 JS specific to be adjacent to the method that is returning this type since there is only one.  The rest will be removed in a future PR once types can be imported.

src/dwn/models/record.js
frankhinek: Sure, why not.  Addressed in 7a40a6622e2eb9fc885d89b83934f373ded6689f

src/dwn/models/record.js
frankhinek: Addressed in 7a40a6622e2eb9fc885d89b83934f373ded6689f

src/did/connect/connect.js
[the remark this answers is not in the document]
dcrousso: but very few of the style changes are actually exposed externally as API.  e.g. why does `parseJSON` need to change?  i doubt that devrel would do any documentation on something like that

src/did/web5-did.js
dcrousso: to be clear, im not opposed to the idea of `web5.did.manager`.  im just commenting on how it is implemented.

for example, instead of
```
clear: (...args) => DidManager.clear(...args, this.didStore),
```
why not just
```
clear: (...args) => this.didStore?.clear()
```

IMO ideally we'd have a `class DIDManager` that's instantiated inside the `constructor` of `Web5DID` instead of creating a new object each time it's called

src/storage/local-storage.js
dcrousso: AFAIK it's only reserved when used on it's own (e.g. `delete foo.bar`) or as an identifier (e.g. `function delete() { }`).  there's nothing disallowing it from being used as a property name (e.g. `Map.prototype.delete`, `Set.prototype.delete`, etc.)

is your concern we'd import something that would try to `export function delete() { }`?  in that case are you expecting that the API surface of that import would _exactly_ match `Storage`?  or would we have a wrapper (e.g. `class FooStorage`)?

src/did/web5-did.js
dcrousso: this was consistent before ca736832e9c3a8b777f7784def4a5f36a27928ca :/

src/dwn/models/record.js
dcrousso: it feels a bit pointless to me (especially since you dont even check the return value when calling it above), so id personally remove it, but i dont feel super strongly

in a more general sense, i dont really see why we need to have exact matching of behavior/structure/etc. between entirely different repositories :/

src/web5.js
dcrousso: my immediate reaction to this was "are you trying to somehow grab a private field from another object?", so I thought it was extremely odd :/

how about we do something ultra verbose like `"web5-message-signing-key"` so it's explicitly clear what it's for (and would allow for other code to explicitly "opt in" to this if they want to customize it for some reason)

src/dwn/interfaces/records.js
frankhinek: Addressed in d08b75f64cfd5a8f21e0e613fecc4477e9bc78e8

src/dwn/models/record.js
frankhinek: Addressed in https://github.com/TBD54566975/web5-js/commit/7a40a6622e2eb9fc885d89b83934f373ded6689f

src/dwn/models/record.js
frankhinek: It isn't mutable directly.  It would be changed as a result of an undelete.

src/dwn/models/record.js
frankhinek: That is a better description.  Originally, the intent was to actually freeze the record in JavaScript.  However, the ability to undelete made that an issue since you cannot unfreeze a frozen record in javascript.

Addressed in d0a95c9f4d5b0c06bcaddea27921b750f9641701

examples/simple-agent/src/utils.js
frankhinek: `INDEX` path was already being set... when defining `messageStore`:

```javascript
const messageStore = new MessageStoreLevel({
  ...
  indexLocation      : `INDEX-${port}`,
});
```

Setting EventLog path addressed in 6d63a0ba443e120f9ffcc0be5020601dbea4f67f

src/did/web5-did.js
frankhinek: > this means that every time someone calls web5.did.manager they're gonna get a new object. can we avoid this? it seems very unnecessary
This doesn't seem to be a major issue given how infrequently this function is called in the normal lifecycle of an application built with Web5 JS.

> also, DidManager seems kinda pointless since all it really does is just call methods on the store. do we really need that collection of functions instead of just this.didStore.thing(arg)?
There are more capabilities coming to DID Manager, including interaction with a Key Store in the future.

> IMO ideally we'd have a class DIDManager that's instantiated inside the constructor of Web5DID instead of creating a new object each time it's called
Little functional difference between these two implementations, but your suggestion is more consistent with how the rest of the Web5 JS lib is implemented.  Refactored in 3f4b8d4a885e2fc59d502621e1a0ac85056536f6

src/dwn/dwn-utils.js
frankhinek: There is ongoing discussion with @mistermoe and @thehenrytsai as to how to handle streams in web/browser vs. Node environments.  This implementation is a temporary workaround until we make a decision on what type of stream (web vs node) to expose in dwn-sdk-js.  Potentially we can move to using the web ReadableStream everywhere but there is more validation to be done to confirm all of the encryption libraries we are using in `dwn-sdk-js` can work with web ReadableStream.

I'll leave this as is for now but open an issue to track making a decision on how to handle the different stream types.

src/dwn/models/record.js
frankhinek: Opened #46 to track considering improvements to the approach

src/dwn/models/record.js
frankhinek: The `self` is necessary as currently written so that the functions within the `dataObj` can access both the instance `this` and the other methods within the object (e.g., `this.text().then(JSON.parse`).

If we refactor the approach then it may be possible but this wouldn't change the functionality.

Opened #46 to track considering improvements to the approach

src/did/web5-did.js
frankhinek: Inconsistency addressed in 1b4d64ec4f93fbcd7785ff2f99dafc6c7c7c6750

src/dwn/models/record.js
frankhinek: Incredibly minor nit but sure...
Addressed in 9786d34e0dedaeb68c519b78844cd613042f819b

src/storage/local-storage.js
frankhinek: Not worth arguing over.  Addressed in 6a95eca9d7dc4f6fe2a6c186ec8d0f85d9fb1f01

src/dwn/interfaces/records.js
frankhinek: Addressed in 96c7a8030cff76db65305578bc3beb34ffc6fc51

src/dwn/interfaces/records.js
frankhinek: Addressed in 96c7a8030cff76db65305578bc3beb34ffc6fc51

src/dwn/interfaces/records.js
frankhinek: Addressed in 96c7a8030cff76db65305578bc3beb34ffc6fc51

src/dwn/interfaces/records.js
frankhinek: Addressed in 96c7a8030cff76db65305578bc3beb34ffc6fc51

src/dwn/models/record.js
frankhinek: Since its a private method, we know and control the type for `propertiesToMutate()`.  However, sure... its more flexible if we ever want to use the method for other purposes.

Addressed in 96c7a8030cff76db65305578bc3beb34ffc6fc51

src/dwn/models/record.js
frankhinek: Addressed in 9870e2fc5d088c1ad216eeecefd0d81b23d0e8f8

src/web5.js
frankhinek: Yes, this was intentional.   The message property would either be present in `response` or created by the conditional just above the highlighted lines.  If it was present in the response, `message` would be an empty object, and therefore, should be overwritten.

However, declaring `message` as an empty object and only conditionally setting properties was not necessary.  I refactored the code in 78d637d1e0646c06fae6ec3b0fd7dbac930bd995

src/transport/http-transport.js
frankhinek: This would not present an issue if you were using the TBD provided libraries for HTTP transport on both the provider and client app.   However, there is the potential for another implementer to accidentally include a `data` property in the `record` object, which would overwrite and cause issues.  To be defensive, I adjusted this in 78d637d1e0646c06fae6ec3b0fd7dbac930bd995

csuwildcat (approved):

PR #60, opened 2023-05-08
chonkfactor

mistermoe:
**NOT COMPLETE!**

too tired right now. will add description tomorrow.


packages/web5/src/dwn-api.ts
```diff
@@ -0,0 +1,169 @@
...
+}
+
+export type ProtocolsQueryRequest = {
+  author: string;
+  message: Omit<ProtocolsQueryOptions, 'authorizationSignatureInput'>
+}
+
+export const DwnApi = (web5Agent: Web5Agent) => ({
+  protocols: {
+    async configure(target: string, request: ProtocolsConfigureRequest) {
```
KendallWeihe, line 63: I see `target` here and I see `tenant` in the `dwn-sdk-js` repo, but is there a distinction with a `did`? As in, a DID is a string (array of bytes, really) of the format defined in [the spec](https://www.w3.org/TR/did-core/), so is `target` and `tenant` some type of superset, or additional functionality of a `did`?

packages/web5/src/dwn-api.ts
```diff
@@ -0,0 +1,169 @@
...
+        target         : target,
+        author         : request.author,
+        messageOptions : request.message,
+        messageType    : DwnInterfaceName.Protocols + DwnMethodName.Query
+      });
+    }
+  },
+  records: {
+    async write(target: string, request: RecordsWriteRequest): Promise<RecordsWriteResponse> {
+      const { author, data, message: requestMessage } = request;
```
KendallWeihe, line 82: Purely a preferential thing, but IMO deconstruction induces a loss of intent. Any time I see a deconstruction, I become skeptical. 

For example, notice how you have to do `requestMessage` and `responseMessage` rather than `request.message` and `response.message`. You have to do this because you have to differentiate whether the `message` is part of the inbound or outbound. Now consider `author` and `data` -- both of which are part of the inbound but purely by looking at the variables that's unclear (I have to trace up the code to determine the context in which those things exist/originated-from).

It's one of those nitpicks, but I think it's worth nitpicking. Consider steel-manning the counter argument, which would go something like, "well `request.data` is verbose, and if I'm using `data` all over the place then I'm going to have a bunch of verbose code." While that's a good point, to that I would respond, "then your function is too large and you may need to introduce more levels of abstraction." That's not the case here, because the function is small enough that verbosity is beneficial for the purposes of clearly conveying intent.

packages/web5/src/dwn-api.ts
```diff
@@ -0,0 +1,169 @@
...
+        ...responseMessage as RecordsWriteMessage,
+      };
+
+      const record = new Record(web5Agent, recordOptions);
+
+      return { ...reply, record };
+    },
+
+    async create(_target: string, _request: RecordsWriteRequest): Promise<RecordsWriteResponse> {
+      // TODO: damn i guess we have to turn this into a class.
```
KendallWeihe, line 124: Why? Is there some sort of encapsulation requirement for creating records? Such as the need for instance properties or methods?

packages/web5/src/dwn-api.ts
```diff
@@ -0,0 +1,169 @@
...
+  ProtocolsQueryOptions,
+  ProtocolsConfigureOptions,
+} from '@tbd54566975/dwn-sdk-js';
+
+import { DwnInterfaceName, DwnMethodName, DataStream } from '@tbd54566975/dwn-sdk-js';
+
+import { Record } from './record.js';
+import { dataToBytes } from './utils.js';
+
+export type RecordsWriteRequest = {
```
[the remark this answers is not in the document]
KendallWeihe: So actually... more thoughts... "SDK" to me implies something heavier than a library. 

In the context of DWN, a library could be a specific implementation of something from the [spec](https://identity.foundation/decentralized-web-node/spec/). For example, say we wanted to use SQL for data storage, but we wanted the DX to follow the DWN interfaces & standards, then that's a library. 

An SDK, to me, usually offers the comprehensive feature set in order to build a simple "Hello World" app but also administrative, or "meta-layer" features which aren't directly applicable to the end-user experience. For example, say the developer wanted to implement a DWeb App which requires orchestration of many-nodes. Such a use case may not directly be defined by the spec, but is totally necessary to really leverage the value of the underlying standard.

I think about it in terms of AWS (or any of the cloud providers). There is the developer experience of _provisioning_ (or, as I'm putting it, "administrating") cloud resources, and then there is the developer experience of actually _using_ those cloud resources. Both together is an "SDK" but solely the latter is a "Library."

I still hold that building a singular artifact which defines solely the interfaces & types is worth consideration.

Idk there's also the matter of debate around a "reference implementation" and how the only realistic path forward is for everyone to use the reference implementation and not implement their own based on the spec. For example, in the context of Bitcoin, there are many implementations, but to this date, Bitcoin Core still holds something like 99% of all traffic. Because the spec is so sophisticated that building your own invariably leads to production issues caused by trying to reconcile differences. 

Idk, I suppose it's semantic-heavy... but perhaps valuable to us all the same.

These things are emergent. If `dwn-sdk-js` has grown heavy and difficult to move then it makes sense to do things like what we've done here.

packages/web5/src/record.ts
```diff
@@ -0,0 +1,245 @@
...
+  toJSON() {
+    return {
+      author        : this.author,
+      target        : this.target,
+      recordId      : this.id,
+      contextId     : this.contextId,
+      dataFormat    : this.dataFormat,
+      dateCreated   : this.dateCreated,
+      interface     : this.interface,
+      method        : this.method,
```
KendallWeihe, line 217: oh nice, we're already doing the `interface` and `method` naming convention here 😂

packages/web5/src/dwn-api.ts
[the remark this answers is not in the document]
KendallWeihe: Anyways... I don't think anything above changes the course with what you're doing here w/ `web5-js`. I think there is some interesting context in what I've written around the broader DX strategy, and nomenclature specifically.

packages/dids/package.json
```diff
@@ -0,0 +1,61 @@
+{
+  "name": "@tbd54566975/dids",
```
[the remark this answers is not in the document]
mistermoe: good question @KendallWeihe was wondering the same myself. i called it `dids` because there are many did method implementations.

any thoughts @frankhinek ?

packages/web5/src/dwn-api.ts
mistermoe: great q. `create` is supposed to be an alias for `write`. in order for me to call `write` from within `create` i'd have to call the `DwnApi` function again

packages/web5/src/dwn-api.ts
mistermoe: > Purely a preferential thing, but IMO deconstruction induces a loss of intent. Any time I see a deconstruction, I become skeptical.
> 
> For example, notice how you have to do `requestMessage` and `responseMessage` rather than `request.message` and `response.message`. You have to do this because you have to differentiate whether the `message` is part of the inbound or outbound. Now consider `author` and `data` -- both of which are part of the inbound but purely by looking at the variables that's unclear (I have to trace up the code to determine the context in which those things exist/originated-from).

such a good point.

packages/web5/src/dwn-api.ts
frankhinek: The intent behind calling it `target` is in the sense that the methods are all DID-relative with DWN Messaging.  While the DID is represented as a string that can be decoded to bytes, it also is cable of being resolved (see [DID Resolution](https://www.w3.org/TR/did-core/#did-resolution)) to a DID document and the DID document is capable of acting as a routing table through the population of [Service](https://www.w3.org/TR/did-core/#services) entries.

You've likely already read these but for anyone else reading the comments:  More info on how [service endpoints](https://identity.foundation/decentralized-web-node/spec/#service-endpoints) and DID-relative [addressing](https://identity.foundation/decentralized-web-node/spec/#addressing).

 The `target` is the intended destination for which a message is to be routed.  When processed, the target DID is resolved to a specific service endpoint which expresses how the sender of the message is able to communicate with the target entity (person, organization, mediator, etc.).

While it is expected that the target DID specified in a DWN message will be a tenant on a DWeb Node, technically it doesn't HAVE to be a multi-tenancy capable DWN implemented as in `dwn-sdk-js`.  You could even imagine a "bridge" node that exists to relay messages.  On one side its speaking DWeb Messages with DWN nodes and on the other it could be any other protocol (e.g., traditional REST API endpoints, etc.).

PR #62, opened 2023-05-11
Fix connect README links

amika-sq:
Noticed that a few of the links in the `web5.connect([options])` section didn't work. Updating to fix!


frankhinek:
Hey @amika-sq -- thanks for the updates.

There was a massive refactor in 0.7.0 that includes so many changes that a lot of the docs will have to be updated.

PR #63, opened 2023-05-12
npm publish action for monorepo

michaelneale:
hey @mistermoe here is the publish action (I removed the snapshot one for now) - this is technically a "monorepo" apparently in npm terms. 

Also for this to publish I think we have to great the npm repos under the tbd... org when we are ready. But can try at any point to create a release.


frankhinek:
Used the NPM workspaces feature rather than handling with bash:
https://github.com/TBD54566975/web5-js/blob/main/.github/workflows/release-npm.yml

PR #64, opened 2023-05-16
dynamic DWN host selection

mistermoe:


mistermoe:
tests fail because tests point to localhost dwn servers

packages/web5/src/web5.ts
```diff
@@ -66,9 +68,20 @@ export class Web5 {
...
 
     if (!profile) {
-      // TODO: add good samaritan nodes here when they're ready
-
-      const defaultProfileDid = await this.did.create('ion');
+      const dwnUrls = await Web5.getDwnHosts();
+      let ionCreateOptions;
+
+      if (dwnUrls.length > 0) {
+        ionCreateOptions = {
```
frankhinek, line 75: To enable us to use show how selecting a `keyId` for message authorization, message attestation, and record encryption work, we need the ION DIDs auto-generated to include the appropriate values as in these test fixtures:

https://github.com/TBD54566975/web5-js/blob/561f475ecf5be74e4f3cf8b530e7bc892af24860/packages/web5/tests/fixtures/test-profiles.ts#L75-L77

https://github.com/TBD54566975/web5-js/blob/561f475ecf5be74e4f3cf8b530e7bc892af24860/packages/web5/tests/fixtures/test-profiles.ts#L26-L47

frankhinek (changes requested):
Need to ensure that all ION DIDs include message authorization, message attestation, and record encryption keys.

packages/web5/src/web5.ts
mistermoe: wondering if this should live [here](https://github.com/TBD54566975/web5-js/blob/main/packages/dids/src/did-ion.ts#L38-L129)

mistermoe: @frankhinek i went ahead and took a stab at including these. decided not to pollute the `DidIonApi.create` method with dwn specific stuff, and instead surfaced a helper method. feel free to change anything/everything about this

packages/web5/src/web5.ts
```diff
@@ -84,4 +91,26 @@ export class Web5 {
 
     return { web5, did: connectedDid };
   }
+
+  /**
```
michaelneale, line 95: this is super cool. 

One suggestion (perhaps future) is to check if the DWN returns 200 OK: 

`
To enhance the code to check that each URL returns a 200 OK status before using it, you can modify the getDwnHosts function as follows:

javascript
Copy code
/**
 * Dynamically selects up to 4 dwn hosts
 */
static async getDwnHosts() {
  const response = await fetch('https://dwn.tbddev.org/.well-known/did.json');

  const didDoc = await response.json();
  const [service] = didUtils.getServices(didDoc, { id: '#dwn', type: 'DecentralizedWebNode' });
  const { nodes } = <DwnServiceEndpoint>service.serviceEndpoint;

  // Allocate up to 4 nodes for a user.
  const numNodesToAllocate = Math.min(Math.floor(nodes.length / 2), 4);
  const dwnUrls = [];

  for (let i = 0; i < numNodesToAllocate; i += 1) {
    let nodeIdx = getRandomInt(0, nodes.length);
    let node = nodes[nodeIdx];
    let nodeResponse = await fetch(node);

    while (nodeResponse.status !== 200) {
      // If the response status is not 200, get another random node
      nodeIdx = getRandomInt(0, nodes.length);
      node = nodes[nodeIdx];
      nodeResponse = await fetch(node);
    }

    dwnUrls.push(node);
  }

  return dwnUrls;
}`

michaelneale: this way should a DWN not be available it will "naturally" load balance.

mistermoe: Ah yeah great idea @michaelneale. I'll add this in now

frankhinek (approved):
LGTM

PR #67, opened 2023-05-17
Fix CI WebKit Tests

frankhinek:
Runs the browser tests using a macOS runner to ensure that Chromium, Firefox, and WebKit tests can all be executed.

### DIFF
--- a/.github/workflows/tests-ci.yml
+++ b/.github/workflows/tests-ci.yml
@@ -37,7 +37,8 @@ jobs:
         run: npm run test:node --ws
 
   test-on-browsers:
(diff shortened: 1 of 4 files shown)


michaelneale:
nice - any reason why it is `macos-latest`?

frankhinek:
> nice - any reason why it is `macos-latest`?

Why not?  “latest” for GitHub runners is latest stable.  So macos-latest is macOS 12 and not 13.   And we use ubuntu-latest in every other test so why deviate?

michaelneale:
@frankhinek oh it was macos instead of ubuntu that confused me (not latest)

PR #75, opened 2023-05-20
Add support for encrypting records using web5.dwn.records.create()

frankhinek:
This PR adds support for encrypting records using the encryption feature in DWN SDK.

The syntax for encrypting a record such that only the local active DID's keys can decrypt:
```typescript
const { record } = await web5.dwn.records.create({
  data: 'Hello',
  encrypt: true,
  message: {
    'schema': 'http://schemas.xyz/message'
  }
});
```

The syntax for encrypting a record such that only `bobDid`'s keys can decrypt:
```typescript
const { record } = await web5.dwn.records.create({
  data: 'Hello',
  encrypt: { for: bobDid },
  message: {
    'schema': 'http://schemas.xyz/message'
  }
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ console.log(records); // an array of record entries from Bob's DWeb Nodes
 
 The query `request` contains the following properties:
 
-- **`from`** - _`DID string`_ (_optional_): the decentralized identifier of the DID of the DWeb Node the query will fetch results from.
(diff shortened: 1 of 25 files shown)


csuwildcat (approved):

frankhinek:
Will submit a new PR that includes encryption and decryption.

PR #88, opened 2023-05-23
Sync Fixes

mistermoe:
* ignore 409s. 409s occur when the record already exists which means the record was already pushed or pulled
* fix `pull` so that it enqueues first.


frankhinek (approved):
LGTM

Issue #82, opened 2023-05-22
messaging test: protocolPath required as well as protocol


michaelneale:
Trying to setup a simple protocol like: 

```
    var simpleMessageProtocol = {
      "protocol": "http://simple-messaging-protocol.xyz",
      "types": {
        "message": {
          "schema": "http://simple-messaging-protocol.xyz/schemas/message",
          "dataFormats": ["text/plain"]
        },
        "messageResponse": {
          "schema": "http://simple-messaging-protocol.xyz/schemas/message-response",
          "dataFormats": ["text/plain"]
        }
      },
      "structure": {
        "message": {
          "$actions": [
            {
              "who": "anyone",
              "can": "write"
            }
          ],
          "messageResponse": {
            "$actions": [
              {
                "who": "recipient",
                "of": "message",
                "can": "write"
              }
            ]
          }
        }
      }
    };

    var { protocol, status } = await web5.dwn.protocols.configure({
      message: {
        definition: simpleMessageProtocol
      }
    });
```

and trying to use that protocol: 

```
    const { record } = await web5.dwn.records.create({
      to: client2,
      data: "Hello World!",
      message: {
        protocol: "http://simple-messaging-protocol.xyz",
        schema: "mic/sample"
      },
    });
    var { status } = await record.send(client2);
    console.log(status);
```

I get an error: 
`records-write.ts:148 Uncaught Error: `protocol` and `protocolPath` must both be defined or undefined at the same time`

What is a protocolPath?

michaelneale:
I was able to look at code and see it would be message (would be nice to document that somewhere). 

I then simplified my protocol: 

```
    var simpleMessageProtocol = {
      "protocol": "http://simple-messaging-protocol3.xyz",
      "types": {
        "message": {
          "schema": "http://simple-messaging-protocol3.xyz/schemas/message",
          "dataFormats": ["text/plain"]
        },
        "messageResponse": {
          "schema": "http://simple-messaging-protocol3.xyz/schemas/message-response",
          "dataFormats": ["text/plain"]
        }
      },
      "structure": {
        "message": {
          "$actions": [
            {
              "who": "anyone",
              "can": "write"
            },
            {
              "who": "recipient",
              "of": "message",
              "can": "read"
            }
          ],
        }
      }
    };
```

And was able to send from one user: 

    const { record } = await web5.dwn.records.create({
      data: "Hello World!",
      message: {
        protocol: "http://simple-messaging-protocol3.xyz",
[...]

PR #89, opened 2023-05-23
add protocolPath and protocol write example

michaelneale:
This was missing from the README

fixes: https://github.com/TBD54566975/web5-js/issues/82


michaelneale:
closing as the reference on the website is better and provides more context

Issue #86, opened 2023-05-22
Make hosted DWNs and sync configurable


frankhinek:
Making the hosted DWNs  was a planned fast-follow to the 0.7.x release, but adding an issue to track since it was requested by multiple community members (example below).

<img width="739" alt="Screenshot 2023-05-22 at 1 19 10 PM" src="https://github.com/TBD54566975/web5-js/assets/134693/10c9cef9-cd3f-489b-9c59-2228f8d51634">

michaelneale:
I am taking a look at this.

michaelneale:
`const { web5, did: myDid } = await connect({ 'dwnEndpoints': ['https://dwn.tbddev.org/dwn0'] });`

frankhinek:
Modifying this issue to only concern making hosted DWNs configurable.  New issue #95 created to track allowing sync behavior to be configurable.

PR #93, opened 2023-05-25
allow overriding of DWN hosts

michaelneale:
adds this enhancement: https://github.com/TBD54566975/web5-js/issues/86

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,22 +50,33 @@ or
 import { Web5 } from CDN_LINK_HERE;
 ```
 
-### **`Web5.connect()`**
+### **`Web5.connect(options)`**
 
 Enables an app to request connection to a user's local identity app (like a desktop or mobile agent - work is underway for reference apps of each), or generate an in-app DID to represent the user (e.g. if the user does not have an identity app).
 
-> **NOTE:** The outputs of this method invocation with be used throughout the other API methods below.
+> **NOTE:** The outputs of this method invocation will be used throughout the other API methods below.
 
 ```javascript
 const { web5, did: myDid } = await Web5.connect();
 ```
 
-connect() takes in an optional `options` object with the following properties: `dwnEndpoints` which is a list of DWeb Node endpoints to connect to if a new DID is to be created. If no `dwnEndpoints` are provided, the SDK will default to built in bootstrap nodes.
+#### **`options`** _(optional)_
 
-For example:
-```javascript
-const { web5, did: myDid } = await connect({ 'dwnEndpoints': ['https://dwn.tbddev.org/dwn0'] });
-```
+An object which may specify any of the following properties:
+
+- **`techPreview`** - _`object`_ _(optional)_: an object that specifies configuration parameters that are relevant during the Tech Preview period of Web5 JS and may be deprecated in the future with advance notice.
+
+  - **`dwnEndpoints`** - _`array`_ _(optional)_: a list of DWeb Node endpoints to define in the DID created and returned by `Web5.connect()`. If this property is omitted, during the Tech Preview two nodes will be included by default (e.g., `['https://dwn.tbddev.org/dwn0', 'https://dwn.tbddev.org/dwn3']`).
+
+  For example:
+
+  ```typescript
+  const { web5, did: myDid } = await Web5.connect({
+    techPreview: {
+      dwnEndpoints: ["https://dwn.your-domain.org/"],
+    },
+  });
+  ```
 
 <!-- > NOTE: This method **_MUST_** be invoked within the scope of a 'trusted user action' (something enforced by the OS/browser) if the desire is to connect to a local identity app. For browsers this is generally some direct user action, like clicking a link or button. -->
 
@@ -86,7 +97,7 @@ Each `Record` instance has the following instance properties: `id`, `attestation
 
 Each `Record` instance has the following instance methods:
 
-- **`data`** - _`object`_: a object with the following convenience methods that read out the data of the record entry in the following formats:
+- **`data`** - _`object`_: an object with the following convenience methods that read out the data of the record entry in the following formats:
   - **`text`** - _`function`_: produces a textual representation of the data.
   - **`json`** - _`function`_: if the value is JSON data, this method will return a parsed JSON object.
   - **`stream`** - _`function`_: returns the raw stream of bytes for the data.
--- a/README.md
+++ b/README.md
@@ -60,11 +60,11 @@ Enables an app to request connection to a user's local identity app (like a desk
 const { web5, did: myDid } = await Web5.connect();
 ```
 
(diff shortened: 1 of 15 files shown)


csuwildcat:
Would rather call the property dwnEndpoints, as host makes it seem like an external entity is always involved, when it could simply be a static IP to a box in your own home.

michaelneale:
@csuwildcat yeah I agree - was struggling for a name. endpoints is also what DID spec calls them, will change.

michaelneale:
@csuwildcat how does this look?

csuwildcat:
Lgtm

frankhinek:
@mistermoe were you going to push up some additions to the PR we discussed yesterday to make overriding other properties possible?  Or should we start with `dwnEndpoints` and open a separate PR?

mistermoe:
i didn't get a chance to add those additional overrides we discussed yesterday. let's start with `dwnEndpoints`. i'll open a separate PR that introduce the other options

mistermoe:
the only thing that gives me pause at the moment is that this `dwnEndpoints` override option only makes sense in a scenario where an embedded user agent is being used. That said, the only agent that exists today is an embedded agent haha! the long term goal is for desktop & mobile agents to exist and it's in those external agents where you'd be providing your custom dwn endpoints or any other custom did configurations. that wouldn't happen via the `web5` api.

Given that we haven't yet finished the first external agent, i think there's value in allowing people to provide custom dwn endpoints today. I think we could do it like so:

```javascript
const { web5, did } = await Web5.connect({
  techPreview: {
    dwnEndpoints: []
 }
});
```

namespacing the `dwnEndpoints` option within a `techPreview` object hopefully makes it more clear that the option is temporary

README.md
```diff
@@ -60,6 +60,13 @@ Enables an app to request connection to a user's local identity app (like a desk
 const { web5, did: myDid } = await Web5.connect();
 ```
 
+connect() takes in an optional `options` object with the following properties: `dwnHosts` which is a list of DWeb Node endpoints to connect to if a new DID is to be created. If no `dwnHosts` are provided, the SDK will default to built in bootstrap nodes.
+
+For example:
+```javascript
+const { web5, did: myDid } = await connect({ 'dwnHosts': ['https://dwn.tbddev.org/dwn0'] });
```
frankhinek: Should be `Web5.connect()` and not `connect()`

frankhinek:
Spotted a couple issues with the existing PR:
- only 1 of the 2 README files was updated
- README example was `const { web5, did: myDid } = await connect(...` instead of `const { web5, did: myDid } = await Web5.connect(...`.

both resolved with Commit [05e4e34](https://github.com/TBD54566975/web5-js/pull/93/commits/05e4e34c0535dda087913e20ea1a294fb2013312)

Still one remaining issue which is if `dwnEndpoints` is an empty array, the `messageAuthorizationKeys` and `recordEncryptionKeys` values aren't populated in the DID document:

with `{ techPreview: { dwnEndpoints: [] } }`:

Screenshot 2023-05-26 at 12 52 46 PM

with `{ techPreview: { dwnEndpoints: ['https://dwn.your-domain.org'] } }`:

Screenshot 2023-05-26 at 12 55 53 PM

Discussing with @mistermoe and @csuwildcat as to how we should handle.

packages/dids/tests/tech-preview.spec.ts
```diff
@@ -0,0 +1,68 @@
+import type { DwnServiceEndpoint } from '../src/types.js';
+
+import { expect } from 'chai';
+
+import { DidIonApi } from '../src/did-ion.js';
+
+describe('Tech Preview', function () {
+  describe('generateDwnConfiguration()', () => {
+    it.only('returns keys and services with two DWN URLs', async () => {
```
diehuxx, line 9: what's with the `it.only`s? are you intentionally trying to exclude other tests from the `package/dids/test` suite? If so, can you use `.skip` on the ones you want to skip instead?

frankhinek: Thanks for catching.  Should have been removed.

packages/web5/src/web5.ts
```diff
@@ -110,20 +112,25 @@ export class Web5 {
   }
 
   /**
-   * dynamically selects up to 4 dwn hosts
+   * Dynamically selects up to 2 DWN endpoints that are provided
+   * by default during the Tech Preview period.
    */
-  static async getDwnHosts() {
+  static async getTechPreviewDwnEndpoints(): Promise<string[]> {
     const response = await fetch('https://dwn.tbddev.org/.well-known/did.json');
```
mistermoe, line 119:
```suggestion
    let response;

    try {
      response = await fetch('https://dwn.tbddev.org/.well-known/did.json');
    } catch(e) {
      console.warn('failed to get tech preview dwn endpoints', e);
      return [];
    }
```

packages/web5/src/web5.ts
```diff
@@ -110,20 +112,25 @@ export class Web5 {
...
 
     // allocate up to 2 nodes for a user.
-    const numNodesToAllocate = Math.min(Math.floor(nodes.length / 2), 2);
-    const dwnUrls = new Set([]);
+    const dwnUrls = new Set<string>();
+    let attempts = 0;
+    const numNodesToAllocate = Math.min(nodes.length, 2);
 
-    for (let i = 0; i < numNodesToAllocate; i += 1) {
+    while(dwnUrls.size < numNodesToAllocate && attempts < nodes.length) {
```
mistermoe, line 133: thoughts on the following?

```typescript
// allocate up to 2 nodes for a user.
const dwnUrls = new Set<string>();
const numNodesToAllocate = Math.min(nodes.length, 2);

for (let attempts = 0; attempts < nodes.length && dwnUrls.size < numNodesToAllocate; attempts += 1) {
  const nodeIdx = getRandomInt(0, nodes.length);
  const dwnUrl = nodes[nodeIdx];

  try {
    const healthCheck = await fetch(`${dwnUrl}/health`);
    if (healthCheck.status === 200) {
      dwnUrls.add(dwnUrl);
    }
  } catch(e: unknown) {
    // Ignore healthcheck failures and try the next node.
  }
}
```

smushes `attempts`

michaelneale:
yeah calling connect with dwnEndpoints specified is in some ways assuming that is the special case that a did will be created for you, but wasn't sure where else it belongs in the short term.

packages/web5/src/web5.ts
frankhinek: Incorporated into latest commit.

packages/web5/src/web5.ts
frankhinek: kinda doesn't smush `attempts` since its still there in the `for` loop 😄 
but LGTM

mistermoe (approved):
🥇

PR #100, opened 2023-05-30
Swappable abstract-level storage 

pax-k:
- added `packages/storage`
- added tests showcasing `memory-level` as swappable storage. Could be used with [`@hyphen-id/rn-async-storage-level`](https://www.npmjs.com/package/@hyphen-id/rn-async-storage-level) once there is react-native support for `dwn-sdk-js` https://github.com/TBD54566975/dwn-sdk-js/issues/309
- if no custom storage is defined, it defaults to `level`
- custom storage can be injected for `appStorage`, `did`, `profileIndex`, `profileStore`, `syncApi` or to `Web5.connect()`

TODO:
- realised that `dwn-sdk-js` should also be refactored to support swappable storage for `eventLog`, `messageStore`, `dataStore`
- will remove `packages/storage` and will move it to `dwn-sdk-js`. web5-js will import the types and `LevelFactory` from it instead

What do you think?

cc @simonwh

### DIFF
--- a/packages/dids/README.md
+++ b/packages/dids/README.md
@@ -1,4 +1,1 @@
 # Web5 DID
-
-Code Coverage
-![Statements](https://img.shields.io/badge/statements-51.04%25-red.svg?style=flat) ![Branches](https://img.shields.io/badge/branches-60%25-red.svg?style=flat) ![Functions](https://img.shields.io/badge/functions-31.25%25-red.svg?style=flat) ![Lines](https://img.shields.io/badge/lines-51.04%25-red.svg?style=flat)


simonwh:
Awesome @pax-k, a big step towards getting `dwn-sdk-js` + `web5.js` running on React Native! 🔥 Hope we can get this (in some version) merged into the main TBD libraries so everyone can benefit 👍

mistermoe:
this is awesome @pax-k. hoping to take a look on friday. 

re moving to `dwn-sdk-js`. we're actually in the process of considering moving a lot of stuff out of `dwn-sdk-js` and into respective packages in this repo (e.g. did & crypto related stuff). i think it probably makes sense for abstract-level storage to live here and we can pull it in as a dependency in `dwn-sdk-js`

pax-k:
Hey @mistermoe, thanks for the reply!

Glad to hear, so actually you consider keeping `packages/storage` here. Just a thought, be aware of circular dependencies in the future: `web5-js` imports `dwn-sdk-js`, and `dwn-sdk-js` will import `web5-js/storage`, but in this case I think there won't be a problem.

Btw I re-opened the PR here https://github.com/TBD54566975/web5-js/pull/104 (changed our branch)

Keep us in the loop 🙌🏻 

Regards

PR #103, opened 2023-06-01
Sync fixes

mistermoe:
This PR fixes two issues related to sync (specifically `push`):

1. account for how record updates work. 
    * updating a record actually creates another `RecordsWrite` with the same recordId. only the first and most recent `RecordsWrite` messages are kept for a given `recordId`. any in between are outright nuked from everywhere. Prior to this PR, sync would attempt to `push` these non-existent records which would end up resulting in `Sync failed due to error:  Error: TODO: message not found`
2. account for scenario where a record is written and deleted _before_ a `push` occurs. 
    * `RecordsDelete` messages keep the initial `RecordsWrite` and simply delete the data associated to that message. Prior to this PR, sync would attempt to fetch data for a deleted record which would end up resulting in `Sync failed due to error:  Error: (404) Failed to read data associated with record <record_id>`

> 💡  Note that https://github.com/TBD54566975/dwn-server/pull/31 needs to be merged in order for the 2nd issue to be fixed

This PR will also be used to bump the relevant package versions to get the fix published the NPMjs.com

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -912,98 +912,6 @@
             "resolved": "packages/dids",
             "link": true
         },
-        "node_modules/@tbd54566975/dwn-sdk-js": {
-            "version": "0.0.33",
-            "resolved": "https://registry.npmjs.org/@tbd54566975/dwn-sdk-js/-/dwn-sdk-js-0.0.33.tgz",
-            "integrity": "sha512-ZdQtRTd0M2VcgGli7kDzkePsuxpwiOg+PRsBJ/UPC5fc6oflzbRqV6Lg9v8bWHozjdnijP58J3RMWmf4cj7bWw==",
-            "dependencies": {
-                "@ipld/dag-cbor": "9.0.0",
-                "@js-temporal/polyfill": "0.4.3",
-                "@noble/ed25519": "1.7.1",
-                "@noble/secp256k1": "1.7.1",
-                "@scure/base": "1.1.1",
-                "@swc/helpers": "0.3.8",
-                "@types/eccrypto": "1.1.3",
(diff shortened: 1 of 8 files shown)


mistermoe:
fix can be tested by doing the following:

1. add the following to `packages/web5/examples`:

```html
<!DOCTYPE html>
<html lang="en">

<head>
  <meta charset="UTF-8">
  <meta http-equiv="X-UA-Compatible" content="IE=edge">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <title>Document</title>
</head>

<body>
  <script type="module">
    import { Web5 } from '/node_modules/@tbd54566975/web5/dist/browser.mjs';

    const { web5, did: aliceDid } = await Web5.connect({
      techPreview: {
        dwnUrls: ['http://localhost:3000']
      }
    });
    console.log(aliceDid);

    const { record, status } = await web5.dwn.records.create({
      data: "Hello Web5",
      message: {
        dataFormat: 'text/plain',
      },
    });

    console.log('WRITE RESULT', status);

    const readResult = await record.data.text();
    console.log('READ RESULT', readResult);

    const updateResult = await record.update({ data: "Hello, I'm updated!" });
    console.log('UPDATE RESULT', updateResult);

    const deleteResult = await record.delete();
    console.log('DELETE RESULT', updateResult);
  </script>
</body>

</html>
```

2. clone `dwn-server`. checkout `tombstones` branch. `npm install && npm run server`
3. run `npx http-server .` from root of `web5-js` repo. fire up browser and head over to `http://localhost:8080`. load html

michaelneale:
Nice 👍

frankhinek (approved):

PR #106, opened 2023-06-02
export all relevant types

pax-k:
Fixes: 

```
The inferred type of 'createProtocolRecord' cannot be named without a reference to '../../../node_modules/@tbd54566975/web5/dist/types/record.js'. This is likely not portable. A type annotation is necessary.ts(2742)
```

What we are doing: 

```
import type { Web5 } from '@tbd54566975/web5'

async createProtocolRecord() {
...
  const response = await this.web5.dwn.records.create({
        data: recordData,
        message: message,
      })
  return response.record!
}
```

### DIFF
--- a/packages/web5/src/main.ts
+++ b/packages/web5/src/main.ts
@@ -1,1 +1,3 @@
 export * from './web5.js';
+
+export type * from './record.js';
--- a/packages/web5/src/main.ts
+++ b/packages/web5/src/main.ts
@@ -1,3 +1,15 @@
 export * from './web5.js';
 
 export type * from './record.js';
+
+export type * from './app-storage.js';
+
+export type * from './did-api.js';
+
+export type * from './did-resolution-cache.js';
+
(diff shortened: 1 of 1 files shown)


michaelneale:
didn't break anything - only risk is that this may be exposing more of an api that originally (of course) but some of those things may be more likely to change (possibly) - just an FYI. What is the normal typescript thing to do - have external types contained in one file?

pax-k:
Hey! The reason for exporting types is that, for example, I'm writing a wrapper on top of web5-js. So far, I only needed `ProtocolDefinition` and `RecordsWriteRequest`, and with my PR I thought to export all types, just to be sure a future PR won't be needed. What do you think?

michaelneale:
ok with me and makes sense - @mistermoe @frankhinek any thoughts on this?

frankhinek:
Addressed in PR #168

PR #112, opened 2023-06-07
fixing errors in Readme's API doc

angiejones:


michaelneale:
good catch, some confusing ones in there.

csuwildcat (approved):

PR #117, opened 2023-06-08
Remote Records Write

KendallWeihe:
Added the ability to write records to remote DWN's via the `to` property (this is in contrast to the `from` property we use in other methods... open to sticking with `from` but that feels a bit awkward)

### DIFF
--- a/packages/web5/src/dwn-api.ts
+++ b/packages/web5/src/dwn-api.ts
@@ -85,6 +85,7 @@ export type RecordsReadResponse = {
 };
 
 export type RecordsWriteRequest = {
+  to?: string;
   data: unknown;
   message?: Omit<Partial<RecordsWriteOptions>, 'authorizationSignatureInput'>;
   store?: boolean;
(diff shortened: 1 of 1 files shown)


frankhinek:
Hey @KendallWeihe -- thanks for putting the PR.

Happy to chat more about this on Discord or live but there was a LOT of discussion, many iterations, and tradeoffs considered both in terms of the happy path and failure scenarios for the `web5.dwn.records.*` and `web5.dwn.protocols.*` methods.

The `to:` option was part of that debate for `web5.dwn.records.write()`, but in the end, consensus formed around the `records.send(<did>)` approach to create/write records on a DWN accessible over a network boundary.  This was admittedly subjective, but `create()`/`write()` have some unique considerations that don't apply to `query()` and `delete()`, which is part of the motivation for having a different interface.

There were 15 different permutations considered for your own local/agent / your own remote DWNs / remote DWNs of others plus the associated failure scenarios.  The (_likely_) most common are listed here: https://github.com/TBD54566975/web5-js/releases/tag/v0.7.0#dwn-ops-cheat-sheet

Given the long list of other high priorities we need to add to the Web5 SDK, this isn't an API we want to introduce changes to just yet until we've got more developer feedback based on building web apps, but we can and should given a running list so that when we've some data to evaluate.

Created Iissue #119 to track so that we can revisit and consider introducing API changes in the future.

PR #120, opened 2023-06-09
Fix eslint NPM scripts and migrate to root .eslintrc config

frankhinek:
This PR will:

- Correct `lint` and `lint:fix` NPM scripts to use `.ts` instead of `.js` since project was refactored to TypeScript:
    https://github.com/TBD54566975/web5-js/blob/5a8e8e3ddc94a6908fe840de6fbfc763ec33f62d/packages/web5/package.json#L11-L12

- Migrate to a single root `.eslintrc.cjs` since all 6 of the existing packages had identical ESLint config files.  If different rules are needed in the future, we can always add a project specific config that extends and modifies the configuration.

- Added `eslint.workingDirectories` setting to VSCode Workspace which is recommended by MSFT to ensure VSCode knows the appropriate CWD to run ESLint from.


mistermoe (approved):

.eslintrc.cjs
```diff
@@ -12,6 +13,10 @@ module.exports = {
     es2022  : true,
     browser : true
   },
+  'ignorePatterns': [
```
mistermoe, line 16: nice find as a replaceent for `.eslintignore`

PR #125, opened 2023-06-13
Adding builder for VerifiableCredential

nitro-neal:
Added VerifiableCredential builder along with unit tests.

The default constructor sets only things that are not optional:
```
    this.verifiableCredential = {
      '@context'        : contexts as ['https://www.w3.org/2018/credentials/v1', ...string[]],
      credentialSubject : credentialSubject,
      issuer            : issuer,
      type              : types,
      issuanceDate      : getCurrentTimestamp(),
    };
```

Added the builder pattern for the rest of the fields with simple validation
Added unit tests
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7793,6 +7793,7 @@
             }
         },
         "packages/credentials": {
+            "name": "@tbd54566975/credentials",
             "version": "0.1.4",
             "license": "Apache-2.0",
             "devDependencies": {
--- a/packages/credentials/src/types.ts
+++ b/packages/credentials/src/types.ts
@@ -4,14 +4,14 @@
  */
 export type VerifiableCredential = {
   /** see {@link https://www.w3.org/TR/vc-data-model/#contexts | Contexts} */
-  '@context': ['https://www.w3.org/2018/credentials/v1', ...string[]];
+  '@context': string[]
   credentialSubject: CredentialSubject | CredentialSubject[];
   /** @see {@link https://www.w3.org/TR/vc-data-model/#identifiers | Identifiers} */
   id?: string;
   /** @see {@link IssuerType} */
   issuer: Issuer;
   /** @see {@link https://www.w3.org/TR/vc-data-model/#types | Types} */
-  type?: string[] | string;
+  type: string[]
   /** @see {@link https://www.w3.org/TR/vc-data-model/#issuance-date | Issuance Date} */
   issuanceDate: string;
   /** @see {@link https://www.w3.org/TR/vc-data-model/#expiration | Expiration} */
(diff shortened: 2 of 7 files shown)


frankhinek (changes requested):
Awesome to see this contribution @nitro-neal !

Looking forward to many more as you help us build out VC capabilities.

frankhinek:
@mistermoe When reviewing this PR, I spotted a few things or had a few questions that I should have probably raised in the initial PR.  I didn't notice them until going through a more thorough review today:

### 1. `VerifiableCredential` type definition - `@context`

Including a static string in the type definition:
https://github.com/TBD54566975/web5-js/blob/d1e2d1003a46542e7c06bfd90156a830933eeef0/packages/credentials/src/types.ts#L5-L7

results in awkward type issues in the code, for example:
https://github.com/TBD54566975/web5-js/blob/f78af053f6775b335bd58cd05d4fe25e76b16736/packages/credentials/src/verifiable-credential-builder.ts#L26-L28

Proposal is to define `@context` as its own type and then reference in the `VerifiableCredential` type:

```typescript
/**
 * The data type for `@context` properties of credentials, presentations, etc.
 *
 * @see {@link https://www.w3.org/TR/vc-data-model/#contexts | Credential Contexts}
 */
export type CredentialContext = string | Record<string, any> | (string | Record<string, any>)[]

/**
 * @see {@link https://www.w3.org/TR/vc-data-model/#credentials | VC data model}
 */
export type VerifiableCredential = {
  /** @see {@link CredentialContext} */
  '@context': CredentialContext;
  credentialSubject: CredentialSubject | CredentialSubject[];
...
  ```

Neil is already including the mandatory `'https://www.w3.org/2018/credentials/v1'` value when a new credential is created, and this seems like a better trade-off to the intellisense code completion resulting from including the string in the type def, since its not likely developers are writing credentials by hand in an IDE.

### 2. `VerifiableCredential` type definition - `type`

Based on my read of the spec: https://www.w3.org/TR/vc-data-model/#types

it appears that `type` is a required property for a `VerifiableCredential`:
Screenshot 2023-06-14 at 9 55 33 AM

and that it must be an array of strings for a VerifiableCredential.

If that is accurate, then we should change the type definition from:
https://github.com/TBD54566975/web5-js/blob/f78af053f6775b335bd58cd05d4fe25e76b16736/packages/credentials/src/types.ts#L13-L14
to:
```typescript
  type: string[];
```

This would make the code that modifies types easier to deal with by not having to contend with strings or arrays of strings and the potential to be undefined.

nitro-neal:
@frankhinek 
image

It seems to me context needs to just be an array of objects according to the spec. In practicality though we use it as an array of strings. In the ssi-sdk we just have a new string array:

contexts := []string{VerifiableCredentialsLinkedDataContext}

nitro-neal:
image

Updating based on comments

mistermoe:
@nitro-neal should we close this in favor of #148 ?

Issue #74, opened 2023-05-19
Allow users to BYODid


mistermoe:
It should be possible to use/import an ION DID created externally when calling `Web5.connect()`.

> **Note**: related to #29

PR #127, opened 2023-06-14
export and then add profile on connect

michaelneale:
Addresses https://github.com/TBD54566975/web5-js/issues/74


pax-k:
What's the status on this?

michaelneale:
@pax-k I think @frankhinek had a few things to tighten up before this, including a different connect flow which would make this break in future (I think I have that right?) cc @csuwildcat

frankhinek:
Addressed in PR #168

PR #134, opened 2023-06-23
Updated ion-tools and dwn-sdk-js dependencies to latest version

thehenrytsai:
Debugged the couple of failed tests and it traced down to incompatibility introduced in protocols in `dwn-sdk-js` currently deployedd. Will need to update the `dwn-server` and get it deployed tomorrow.


mistermoe (approved):
best

frankhinek:
@thehenrytsai if you rebase on main, you can pull in the fix in PR #136 for the `semver` security vulnerability that was causing this PR to fail the security audit check.

PR #143, opened 2023-07-05
Add delete functionality to ProfileAPI

amika-sq:
The `ProfileApi`, which implements the `ProfileManager` interface, currently has the ability to:
* Create a Profile
* Get a specific Profile
* List all Profiles

There currently isn't a way to delete a profile. The various agents (desktop & mobile) will allow users full management over their profiles, including the option to delete them.

This PR adds in this delete functionality, and adds a test around the functionality to make sure it works as expected.


frankhinek:
Package and ProfileApi have been removed/refactored.

PR #144, opened 2023-07-05
increase default sync interval from `1s` ->  `2m`

mistermoe:
Per @mistermoe, as reported on Discord:
> If you leave the network tab open for long enough in safari or chrome 
> The tab crashes
> Bc of the polling

2 minute interval was based on discussion on Discord but could be revisited based on new information in the future.


frankhinek (approved):
LGTM

Issue #133, opened 2023-06-22
Bug: Could not find a declaration file for module '@tbd54566975/crypto'


cre8:
I used a fresh installed Angular (v16) and node version 19.9 and wanted to add `@tbd54566975/web5`
But after starting, I get the following errors:
```
Build at: 2023-06-22T12:03:01.773Z - Hash: c7658a62e29cc2dc - Time: 4377ms

Error: node_modules/@tbd54566975/dids/dist/types/did-ion.d.ts:1:50 - error TS7016: Could not find a declaration file for module '@tbd54566975/crypto'. '/home/mimo/Web5/node_modules/@tbd54566975/crypto/dist/cjs/main.cjs' implicitly has an 'any' type.
  Try `npm i --save-dev @types/tbd54566975__crypto` if it exists or add a new declaration (.d.ts) file containing `declare module '@tbd54566975/crypto';`

1 import type { PublicKeyJwk, PrivateKeyJwk } from '@tbd54566975/crypto';
                                                   ~~~~~~~~~~~~~~~~~~~~~


Error: node_modules/@tbd54566975/dids/dist/types/types.d.ts:2:50 - error TS7016: Could not find a declaration file for module '@tbd54566975/crypto'. '/home/mimo/Web5/node_modules/@tbd54566975/crypto/dist/cjs/main.cjs' implicitly has an 'any' type.
  Try `npm i --save-dev @types/tbd54566975__crypto` if it exists or add a new declaration (.d.ts) file containing `declare module '@tbd54566975/crypto';`

2 import type { PublicKeyJwk, PrivateKeyJwk } from '@tbd54566975/crypto';
                                                   ~~~~~~~~~~~~~~~~~~~~~


Error: node_modules/@tbd54566975/dids/dist/types/utils.d.ts:1:33 - error TS7016: Could not find a declaration file for module '@tbd54566975/crypto'. '/home/mimo/Web5/node_modules/@tbd54566975/crypto/dist/cjs/main.cjs' implicitly has an 'any' type.
  Try `npm i --save-dev @types/tbd54566975__crypto` if it exists or add a new declaration (.d.ts) file containing `declare module '@tbd54566975/crypto';`

1 import type { KeyPairJwk } from '@tbd54566975/crypto';
```
To avoid the 
```
[...]

PR #145, opened 2023-07-06
Resolve Issue #133 by removing tests from crypto and proxy-agent tsconfig

frankhinek:
Discovered today that the root cause of Issue #133 is that two of the Web5.js packages differ from the others in terms of the directives in `tsconfig.json`.  The crypto and web5-proxy-agent packages have the tests directory under include:
```json
  "include": [
    "src",
    "tests" // building tests also so we can run tests directly without the need for ts-node, ts-mocha
  ],
```

which results in the `dist/types` directory having sub-directories for `src` and `tests`:
image

instead of the type declarations being directly under `types`:
image

which then breaks the path in package.json for `types`:
image

since the path is actually `./dist/types/src/main.d.ts`.

This PR removes `tests` from being included in the `tsconfig.json` configuration and also removes an errant `typings` exclusion in the `crypto` package since that is only needed in the `dids` package.

### DIFF
--- a/packages/crypto/tsconfig.json
+++ b/packages/crypto/tsconfig.json
@@ -18,8 +18,7 @@
     "esModuleInterop": true
   },
   "include": [
-    "src",
-    "tests" // building tests also so we can run tests directly without the need for ts-node, ts-mocha
+    "src"
   ],
   "exclude": [
     "node_modules",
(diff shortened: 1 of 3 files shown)


mistermoe (approved):
awesome

PR #148, opened 2023-07-07
adding vc create

nitro-neal:
This change allows for a VC to be created with default values. Only the credentialSubject is required to create a credential

- web5.vc.create() - Creates a Self Signed VC and stores it in local DWN
    - `credentialSubject: any` - the json object of the credentialSubject field
    - `options?: CreateOptions` - this will be the options for the other fields that consist of a VerifiableCredential like(“@context”, status, subject, expirationDate, id, issuanceDate, issuer, type, etc.)
    - returns a `Record` of the saved VC and the vc data object itself
    
    
 Note:
  * options is not implemented in this PR.
  * signing is not implemented in this PR.


packages/web5/src/vc-api.ts
```diff
@@ -1,16 +1,79 @@
...
+    };
+
+    // TODO: Sign VC
+    // const signedVc = this.#web5Agent.sign(vc);
+
+    const { record, status } = await this._writeRecord(vc);
+    return { record, status, vc };
+  }
+
+  private async _writeRecord(vc: VerifiableCredential) {
```
nitro-neal, line 47: This is just the impl to records.write()
Is this ok for now or should we:

1. do web5.dwn.recrods.write()?
2. make a util function in dwn.record package that they can both share?

packages/web5/src/vc-api.ts
```diff
@@ -1,16 +1,79 @@
...
+    const messageOptions: Partial<RecordsWriteOptions> = { ...{ schema: 'vc/vc', dataFormat: 'application/json' } };
```
nitro-neal, line 48: Any thoughts on what the schema should be?

packages/web5/src/vc-api.ts
nitro-neal: After thinking about it, I think we should be fine using web5.dwn.records.write.. 
After all this is the web5 api and they should be able to depend on each other, the credentials/ package will be stand along and not have anything like that

packages/web5/src/web5.ts
```diff
@@ -5,7 +5,7 @@ import type { DidState, DidMethodApi, DidResolverCache, DwnServiceEndpoint } fro
 import ms from 'ms';
 
 // import  { Web5ProxyAgent } from '@tbd54566975/web5-proxy-agent';
-import { Dwn } from '@tbd54566975/dwn-sdk-js';
+import { Dwn} from '@tbd54566975/dwn-sdk-js';
```
frankhinek, line 8:
```suggestion
import { Dwn } from '@tbd54566975/dwn-sdk-js';
```

packages/web5-user-agent/src/web5-user-agent.ts
```diff
@@ -280,6 +322,54 @@ export class Web5UserAgent implements Web5Agent {
...
+    const signatureInputBytes = Encoder.stringToBytes(signatureInput);
+
+    const signatureArrayBuffer = await this.keyManager.sign({
+      algorithm : { name: 'ECDSA', hash: 'SHA-256' },
+      keyRef    : keyPair.privateKey.id,
+      data      : signatureInputBytes,
+    });
+
+    let uint8ArraySignature = new Uint8Array(signatureArrayBuffer);
+    const signatureBase64url = Encoder.bytesToBase64Url(uint8ArraySignature);
```
frankhinek, line 368:
```suggestion
    const signatureBase64url = Convert.arrayBuffer(signatureArrayBuffer).toBase64Url();
```

packages/web5-user-agent/src/web5-user-agent.ts
```diff
@@ -190,6 +203,35 @@ export class Web5UserAgent implements Web5Agent {
     };
   }
 
+  async processVcRequest(request: ProcessVcRequest): Promise<VcResponse> {
+    const signedJwt = await this.#sign(request.vc as VerifiableCredential);
+
+    const messageOptions: Partial<RecordsWriteOptions> = { ...{ schema: 'vc/vc', dataFormat: 'application/vc+jwt' } };
+    const { dataBlob, dataFormat } = dataToBlob(signedJwt, messageOptions.dataFormat);
```
frankhinek, line 210: Similar to the approach in `DwnApi` you'd want to get the data into `Blob` format as early as possible (i.e., do this conversion in `VcApi`) since this is the standard format/encoding we're using when data has to be transported over a network boundary using `fetch()`.  It is unlikely that verifiable credentials would be in the dozens or hundreds of MB in data size, but might as well be consistent.

Here's where it's done in `DwnApi`:
https://github.com/TBD54566975/web5-js/blob/main/packages/web5/src/dwn-api.ts#L327-L342

packages/web5-user-agent/src/utils.ts
```diff
@@ -1,10 +1,40 @@
...
 
 export function blobToIsomorphicNodeReadable(blob: Blob): Readable {
   return webReadableToIsomorphicNodeReadable(blob.stream());
 }
 
 export function webReadableToIsomorphicNodeReadable(webReadable: ReadableStream) {
   return new ReadableWebToNodeStream(webReadable);
+}
+
+export function dataToBlob(data, dataFormat){
```
frankhinek, line 13: Per the other comment about moving the `Blob` conversion to `VcApi`, since its already in `packages/web5/src/utils`, you can import from there in the `web5` package and remove this from `web5-user-agent`.

packages/web5-user-agent/src/web5-user-agent.ts
```diff
@@ -280,6 +322,54 @@ export class Web5UserAgent implements Web5Agent {
     return { message: dwnMessage.toJSON(), dataStream: readableStream };
   }
 
+
+  // TODO: have issuer did key already stored in key manager and use that instead of generating a new one
```
frankhinek, line 326: You got it -- you'd want to reference the key that's already stored.

Worth noting that there's a convention we've been using for signing DWN messages and encrypting DWN records, which is noted here:
- https://github.com/TBD54566975/web5-js/blob/d8e2be7380e1ad504e7467d990dd942cead04ee5/packages/dids/src/types.ts#L61-L66
- https://github.com/TBD54566975/web5-js/blob/d8e2be7380e1ad504e7467d990dd942cead04ee5/packages/dids/src/did-ion.ts#L175-L182

To simplify we're going to combine `messageAuthorizationKeys` and `messageAttestationKeys`  to `messageSigningKeys` since the same key can be used for both.  Attestation is an option for a message if the author wants non-repudiable attestation.  For now, you can resolve the DID document of the other, get the service endpoints using [`getServices()`](https://github.com/TBD54566975/web5-js/blob/d8e2be7380e1ad504e7467d990dd942cead04ee5/packages/dids/src/utils.ts#L47-L53) from the `dids` package and use the first index in the `messageAuthorizationKeys` array as the key reference to ask the `KeyManager` to sign with.

frankhinek (changes requested):

packages/web5-agent/src/web5-agent.ts
```diff
@@ -54,6 +58,28 @@ export type DwnResponse = {
...
+
+export type SendVcRequest = {
+  author: string;
+  target: string;
+  vc: VerifiableCredential;
+  messageCid?: string;
+};
+
+export type VcResponse = {
+  vcDataBlob?: Blob;
```
frankhinek, line 78: Does this need to be a Blob?  or can it be JSON?

packages/web5-agent/src/web5-agent.ts
```diff
@@ -54,6 +58,28 @@ export type DwnResponse = {
   reply: UnionMessageReply;
 };
 
+/**
+ * TODO: add JSDoc
+ */
+export type ProcessVcRequest = {
```
frankhinek, line 64: May need to add `dataStream?: ...` here once you move the data to blob conversion into `VcApi`

packages/web5-user-agent/src/web5-user-agent.ts
nitro-neal: I had to blob in this way because of signing:

I need to sign the vcJWT before I turn it into a blob

Should I create a new method on web5Agent: 
web5Agent.sign()?

or should I send the unsigned blob, un blobify it, sign it, then blobify it again?

nitro-neal: well actually if I make web5Agent.sign().. I will need to blob it, send it to sign, unblob it, sign it..

mistermoe:
Merging this will come after @frankhinek 's agent PR

ianpatton:
I implemented this differently in #243

thehenrytsai:

PR #152, opened 2023-07-09
Add Crypto Algorithm, KeyManager, and KMS to crypto package

frankhinek:
This PR adds a significant amount of functionality to the `crypto` package but does so in a way that very minimal changes were made to existing functionality.  Once merged, these capabilities will be incrementally used across the Web5.js SDK.

> **Note**
> There are no changes to any external `web5.*.*` APIs.

### `common` package
- Adds `Convert` utility to convert between data and encoding types.
- Adds `Multicodec` utility to make it easier to add and remove Multicodec prefixes, primarily used in the generation of DID identifiers.
- Adds a `MemoryStore` implementation that can be used for prototyping or testing key/value stores used by any project.
- Adds Type utilities: `isArrayBufferSlice`, `isDefined`,  and `universalTypeOf`

### `crypto` package
- The `crypto` package previously had a dependency on `multiformats` but the inclusion of the functionality that required this dependency was not crypto specific.  As a result, the common functionality was moved to the `common` package in the form of the new `Multicodec` implementation and the DID-specific functionality was moved to the `dids` package in the form of a new `keyToMultibaseId` function.
- Adds crypto algorithms API as a base that can be extended for concrete implementations of block and stream ciphers, digital signature algorithms, etc.
- Adds concrete implementations of crypto algorithms, including AES-CTR, ECDH (secp256k1 & X25519), ECDSA (secp256k1), and EdDSA (Ed25519).
- Adds crypto primitive implementations that can be used to create a variety of crypto systems/suites, including AES-CTR, AES-GCM, ConcatKDF, Ed25519, secp256k1, X25519, and XChaCha20.
- Adds a KeyManager implementation that can be used by Web5 Apps/Agents to provide a common interface to generate, encrypt, decrypt, sign, verify, etc. methods without directly interacting with private key material.  Can support multiple KMS implementations once implemented (e.g., AWS, GCP) but includes a default KMS that can be backed by a key-value store.
- Adds a default KeyManagementSystem implementation that can be backed by a key-value store.
- Adds a verity of utilities, including: `checkValidProperty`, `checkRequiredProperty`, `isBufferKeyPair`, `isCryptoKeyPair`, `isManagedKey`, `isManagedKeyPair`, and `randomUuid`.

### `dids` package
- The `DidKeyApi` implementation previously depended on multicodec and multibase functionality in the `crypto` package that wasn't crypto specific.  To reduce unnecessary dependencies in the `crypto` package, the common functionality was moved to the `common` package in the form of the new `Multicodec` implementation and the DID-specific functionality was moved to the `dids` package in the form of a new `keyToMultibaseId` function.
- As a result of the changes mentioned in the prior bullet, a few minor changes were made the to `DidKeyApi` implementation to import the needed functionality.


packages/common/src/convert.ts
```diff
@@ -0,0 +1,307 @@
+import type { Multibase } from 'multiformats';
+
+import { base58btc } from 'multiformats/bases/base58';
+import { base64url } from 'multiformats/bases/base64';
```
mistermoe, line 4: thoughts on using `paulmillr`'s [`scure-base`](https://github.com/paulmillr/scure-base) for base encoding? makes sense to use multiformats given that we've already pulled it in as a dependency. if we do decide to switch to `scure-base` we'll want to use `base64urlnopad` ([context](https://github.com/paulmillr/scure-base/issues/4))

mistermoe (approved):
so excited

packages/common/src/convert.ts
frankhinek: Great idea -- created an issue to track follow-up https://github.com/TBD54566975/web5-js/issues/153

PR #155, opened 2023-07-12
Temporarily expose access to local ProfileManager through `web5.techPreview.profileManager`

mistermoe:
enables the ability to import/export profiles that will change once future changes land


frankhinek (approved):
Fixed the title since the proposal is to expose the ProfileManager instance and not the user agent.

This approach is not ideal but best we can do until refactored agent approach is ready for PR.  Getting close.

frankhinek:
Addressed in PR #168

Issue #154, opened 2023-07-12
recordId should be a property of Filter


angiejones:
When querying by `recordID`, it's unnecessary to force the developer to also specify filter criteria since there's only one record that can match a record id.

```js
const {records} = await web5.dwn.records.query({
  message: {
    recordID: 'bafyreiaz5oycqbrnmmpvffxqyoqxvx6bcnqueprmt2qnvzcurpc52r5uyy',
    filter: {
      dataFormat: 'text/plain',
    }
  }
});
```

My suggestion is to make `filter` optional and only require it if `recordID` is not supplied.

This is not a major issue but it is a weird developer experience to force me to provide extra criteria when it's not actually needed to find the record by ID.

<hr/>

Update: my new request is to move `recordId` to the `filter` object to adhere to the DIF spec

> I checked the [DIF spec ](https://identity.foundation/decentralized-web-node/spec/#recordsquery)to make sure we're still adhering to it, and it does say that the object _may_ contain `filter` (not must). But it also shows `recordId` as one of the properties of `filter`. So, I think the fix here is just to move `recordId` to the `filter` object.
> 
> ![image](https://user-images.githubusercontent.com/15972783/253831823-8ea16613-419e-49f7-ad8e-4a900881b838.png)

Developerayo:
Love the idea! @angiejones 
And it does ease development processes, 

Developers will no longer need to specify the `filter` object if a `recordID` was specified, and if a `recordID` is provided the `filter` object should equally be optional, and if none we just throw and error. 

Want me to push a PR for this? @angiejones @frankhinek or there's an internal idea around this.

angiejones:
> Love the idea! @angiejones And it does ease development processes,
> 
> Developers will no longer need to specify the `filter` object if a `recordID` was specified, and if a `recordID` is provided the `filter` object should equally be optional, and if none we just throw and error.
> 
> Want me to push a PR for this? @angiejones @frankhinek or there's an internal idea around this.

sure, go for it! thanks Ayo! 🙏🏾

angiejones:
I checked the [DIF spec ](https://identity.foundation/decentralized-web-node/spec/#recordsquery)to make sure we're still adhering to it, and it does say that the object *may* contain `filter` (not must). But it also shows `recordId` as one of the properties of `filter`. So, I think the fix here is just to move `recordId` to the `filter` object.

[...]

PR #157, opened 2023-07-17
Add `recordID` as a property of filter 

Developerayo:
## Overview
This pull request makes the `recordID` property the only required filed for `filter`, simplyfing the developer's experience. 
Also, the `dataFormat` property can still be added but it's now optional.
 
See #154 for more details

## Example

```javascript
const {records} = await web5.dwn.records.query({
  message: {
    filter: {
      recordId: 'bafyreiaz5oycqbrnmmpvffxqyoqxvx6bcnqueprmt2qnvzcurpc52r5uyy',
    }
  }
});
```

### DIFF
--- a/packages/web5/src/dwn-api.ts
+++ b/packages/web5/src/dwn-api.ts
@@ -206,7 +206,7 @@ export class DwnApi {
       /**
        * TODO: Document method.
        */
-      delete: async (request: RecordsDeleteRequest): Promise<RecordsDeleteResponse> => {
+      delete: async (request: { from?: string, message: { filter: { recordId: string } } }): Promise<RecordsDeleteResponse> => {
         const agentRequest = {
           author         : this.connectedDid,
           messageOptions : request.message,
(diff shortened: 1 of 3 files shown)


frankhinek:
@Developerayo see comments here: https://github.com/TBD54566975/web5-js/issues/154#issuecomment-1640066133

The example you provided is already possible today without any changes:

```typescript
const {records} = await web5.dwn.records.query({
  message: {
    filter: {
      recordId: 'bafyreiaz5oycqbrnmmpvffxqyoqxvx6bcnqueprmt2qnvzcurpc52r5uyy',
    }
  }
});
```

Adding a `filter` property and requiring developers to specify the filter for RecordsDelete and RecordsRead isn't desirable as these are not queries and the only property specified is `recordId`.  Such a change would make every `web5.dwn.records.read()` and `web5.dwn.records.delete()` unnecessarily verbose and would be confusing as it would lead to some developers thinking that query filters can be passed to `delete()` and `read()` like they can with `query()`.

Developerayo:
Ah! I see now @frankhinek 
Thanks for weighing in.

Will close this and open up another PR that includes information about all of the possible records query filters as you said in #154

Issue #95, opened 2023-05-28
Make sync configurable


frankhinek:
Sync should be configurable and this feature was requested by multiple community members (example below).

<img width="739" alt="Screenshot 2023-05-22 at 1 19 10 PM" src="https://github.com/TBD54566975/web5-js/assets/134693/10c9cef9-cd3f-489b-9c59-2228f8d51634">

mistermoe:
initial semi-delirious thoughts:

```typescript
const { web5, did } = await Web5.connect({
  techPreview: {
    sync: '5m' | true | false
  }
});
```

* value provided for `sync` can be a `boolean` or a `string`. 
* if string, value is parsed by https://www.npmjs.com/package/ms. 
* if `false`, toggles sync off entirely
* if `true`, toggles sync on with default interval of 1m (totally down for default to be longer)
* defaults to `true`
* the reason i put the option in `techPreview` is because i think configuring sync's interval is actually an agent configuration that wouldn't be controlled by `web5`. imagine alice using multiple web5 apps that are all constantly clobbering s single sync interval for some remote agent
* def lmk if allowing for boolean + hipster string time format feels too "cute".

mistermoe:
another relevant discord discussion starts [here](https://discord.com/channels/937858703112155166/1068273971432280196/1114237954731278388)

PR #204, opened 2023-08-30
Add SyncManager to Web5 Managed Agent

frankhinek:
This PR will:
- Re-introduce [this sync implementation](https://github.com/TBD54566975/web5-js/blob/main/packages/old/web5-user-agent/src/sync-api.ts) to restore sync capability.
- Resolves an issue with the original sync implementation with cascading echoes that would occur due to random lexicographical ordering of sync jobs.  Unresolved is a single echo after a pull operation, which is tracked in Issue #207.
- Resolves Issue #95.
- Previously sync had no tests so basic tests have been added for `push()` and `pull()`.  Testing for more scenarios and edge cases is needed, which is tracked in Issue #206
- Temporarily adds utility methods to `DwnManager` that should be refactored as tracked in Issue #205.
- Adds `SyncManager` to the tests / test agents.
- Improves test coverage in the `@web5/agent` package for `DwnManager` handling `EventsGet`, `MessagesGet`, `ProtocolsConfigure`, and `ProtocolsQuery`.
- Adds test coverage in the `@web5/api` package for `Protocol`.

### DIFF
--- a/packages/agent/src/dwn-manager.ts
+++ b/packages/agent/src/dwn-manager.ts
@@ -532,4 +532,118 @@ export class DwnManager {
 
     return jws;
   }
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+  /**
+   * ADDED TO GET SYNC WORKING
+   * - createMessage()
+   * - processMessage()
+   * - writePrunedRecord()
+   */
+
+  public async createMessage(options: {
+    author: string,
+    messageOptions: unknown,
+    messageType: string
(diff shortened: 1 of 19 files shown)


packages/agent/src/dwn-manager.ts
```diff
@@ -532,4 +532,118 @@ export class DwnManager {
...
+
+
+
+
+
+
+
+
+
+
```
mistermoe, line 560: adding these newlines to create some distance between what we intend to keep around vs. the stuff we hope to refactor?

mistermoe (approved):
🥇

packages/agent/src/dwn-manager.ts
frankhinek: Yes, that's the only reason they are there -- as a reminder to get rid of this patchwork of methods when we refactor.

PR #215, opened 2023-09-12
100% tests coverage for `web5/crypto`

adam4leos:
Screenshot 2023-09-12 at 3 57 46 PM

### DIFF
--- a/packages/crypto/src/jose.ts
+++ b/packages/crypto/src/jose.ts
@@ -863,7 +863,7 @@ export class Jose {
     const jose = multicodecToJoseMapping[lookupKey];
 
     if (jose === undefined) {
-      throw new Error(`Unsupported Multicodec to JOSE conversion: '${options.name ?? options.code}'`);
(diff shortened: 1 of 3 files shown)


packages/crypto/src/jose.ts
```diff
@@ -863,7 +863,7 @@ export class Jose {
     const jose = multicodecToJoseMapping[lookupKey];
 
     if (jose === undefined) {
-      throw new Error(`Unsupported Multicodec to JOSE conversion: '${options.name ?? options.code}'`);
```
adam4leos, line 866: this is unreachable;
we can't hit this line with undefined code AND name, as undefined code would throw in `getNameFromCode` with undefined name

Removed to reach 100% tests coverage, as it's not possible to cover unreachable edge case...

frankhinek: Appreciate the clarification

frankhinek (approved):
LGTM

PR #218, opened 2023-09-14
adding required property 'published' to protocols.configure example

angiejones:
without the `published` property on a protocol definition, the following error occurs:

Error: /descriptor/definition: must have required property 'published'


frankhinek (approved):
Added change to the `packages/api/README.md` to keep the files in sync but otherwise LGTM

PR #222, opened 2023-09-19
Update Sync Interface

LiranCohen:
This is a WIP but opening it up for some input and reviews.

Goals of this PR:
 - to simplify the SyncManager interface and make it easier to test
 - add support for testing multiple remote DWNs
 
 This PR prepares the pull/push queues before calling sync, this will allow to test out different strategies for enqueuing syncs, and currently allows pull/push to be called async after enqueuing.
 
I have some ideas of changes to make to the DWN SDK/DWN Server to make sync more efficient, but will work through those in a separate PR.

Additionally, I am creating some chaos-monkey type scripts that will live in a separate repo. This will allow us to also compare syncing performance with different scenarios.
 
 **** NOTES
I think we should enforce a minimum sync, however for certain tests this minimum should be low. I've added logic for the testing environments, but locally you currently have to pass a specific ENV variable for tests to pass. This is just temporary.

Looking for input on how best to handle this, or just remove the minimum interval from the code completely.


frankhinek:
Talked to @LiranCohen about this PR yesterday and he's going to review and determine whether it should be closed or taken out of Draft state for review.

frankhinek:
@LiranCohen said he'll re-open later once ready for review

PR #224, opened 2023-09-24
Generate VUK with crypto.subtle

leordev:
## Problem
I'm experiencing a slow initialization of the `Web5.connect()` fn, to the point where the browser freezes for ~4-5 seconds.

It can be reproduced here: https://bucolic-cendol-324631.netlify.app/ -- just click in sign in and then press **Use Browser In-App** Agent. You will notice that the loading spinning animation even freezes while the connect is running. After being logged in you can also just refresh the page and you will notice how slow it takes to load the page, this is because if I recognize a session was initiated I just call `Web5.connect()` before rendering the page.

## Fix

I was debugging the `connect()` fn stepping each line and I narrowed it down to the [`AppDataVault#generateVaultUnlockKey()`](https://github.com/TBD54566975/web5-js/blob/main/packages/agent/src/app-data-store.ts#L146-L157).

From the conversation (link below) it seems that `@noble/hashes/pbkdf2` does not use the native browser crypto subtle which runs way faster. So I just did that.

I tested in my web5-music app and it reduces from 5 seconds to less than 2 seconds and it's non blocking now (the browser does not freeze).

## Reference
The whole context about this issue can be found in our Discord #web5 in [this conversation](https://discord.com/channels/937858703112155166/969272658501976117/1153777451013513257)

### DIFF
--- a/packages/agent/src/app-data-store.ts
+++ b/packages/agent/src/app-data-store.ts
@@ -8,6 +8,7 @@ import { sha256 } from '@noble/hashes/sha256';
 import { randomBytes } from '@web5/crypto/utils';
 import { Convert, MemoryStore } from '@web5/common';
 import { CryptoKey, Jose, XChaCha20Poly1305 } from '@web5/crypto';
+import { generateVaultUnlockKey } from './vuk-generator.js';
 
 export type AppDataBackup = {
   /**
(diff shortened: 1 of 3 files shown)


shamilovtim:
Hey @leordev I agree with making as much of our SDK use native functions as possible! 

One thing, many of our crypto libs first check for node crypto, and then if node crypto isn't available, they fallback to web crypto (`crypto.subtle`). This approach is beneficial for runtimes like React Native or node versions less than node 16 which don't currently have great support for `crypto.subtle`. You can look through the insides of our crypto dependencies for how they handle this and follow their lead for how they handle using node crypto first and then dropping back down to subtle crypto as a contingency.

leordev:
Adding our discord thread for more context on the PR [here](https://discord.com/channels/937858703112155166/1155959400029229118).

Great suggestion. I found a few variations of how subtle is instantiated, and this was the closest one to our case:

browserify/pbkdf2 lib - which hints this in [their readme](https://github.com/browserify/pbkdf2):
> For high performance, use the async variant (pbkdf2.pbkdf2), not pbkdf2.pbkdf2Sync, this variant has the oppurtunity to use window.crypto.subtle when browserified.

Check it out: https://github.com/browserify/pbkdf2/blob/master/lib/async.js#L113-L114

Which is exactly what we are trying to do here but falling back to noble, which is audited... I will add this change in my pr!

leordev:
So as we aligned in our sync I'm checking for subtle crypto first and then falling back to noble here:

https://github.com/TBD54566975/web5-js/blob/1b641f99f624c9d9713ae61f3a5f1886929c3bf8/packages/agent/src/app-data-store.ts#L147-L151

I was going through the path of dynamically importing `node:crypto` and using `pbkdf2()` but that seemed unnecessary just because of the polyfill that we require for node <=18:

```js
// node.js 18 and earlier,  needs globalThis.crypto polyfill
import { webcrypto } from "node:crypto";
// @ts-ignore
if (!globalThis.crypto) globalThis.crypto = webcrypto;
```

We will always have the subtle in node because of this polyfill...

shamilovtim:
I think we still want node crypto as an option (non subtle). Could it be included here?

shamilovtim:
What's the benefit of falling back to noble after node crypto and web crypto? Are there runtimes with no native crypto library at all?

leordev:
I've added the fallback to `node:crypto` and removed the noble one. If it makes sense please let me know and I will work on the unit tests!

shamilovtim (approved):

frankhinek:
Implemented as part of `@web5/crypto` in PR #262

PR #227, opened 2023-09-27
Export tech preview

amika-sq:
Mobile currently cannot exercise sync capabilities, as the generated managed identities do not have DWN endpoints associated with them.

Exporting the functions from the `tech-preview` file will allow mobile to call `getTechPreviewDwnEndpoints` to acquire the test endpoints, and associate them with created ION DIDs. This will allow mobile to test if sync works.

### DIFF
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -5,3 +5,4 @@ export * from './record.js';
 export * as utils from './utils.js';
 export * from './vc-api.js';
 export * from './web5.js';
+export * from './tech-preview.js';


frankhinek (approved):
Hey @amika-sq -- looks good

You probably already found this, but in case its useful here's the code to use `getTechPreviewDwnEndpoints()` with `identityManager.create()`:

https://github.com/TBD54566975/web5-js/blob/a5729b60c03625e6e7daa2b653c8d95bb21b3dc3/packages/api/src/web5.ts#L104-L113

PR #229, opened 2023-09-28
sendDwnRequest returning web Readable

LiranCohen:
`DWNManager` has `sendRequest` and `processRequest` both return the `DwnResponse` type.

However `sendRequest` returns a web Readable as the `data` property while `processRequest` returns an isomorphic Readable.

This was causing confusion for me during testing as dealing with the data is different between the two.

I added some tests that around large data payloads in order to test data streams.


frankhinek:
Talked to @LiranCohen about this PR yesterday and he's going to review and determine whether it should be closed or taken out of Draft state for review.

frankhinek:
@LiranCohen said he'll re-open later once ready for review

PR #236, opened 2023-10-03
Remove old package

amika-sq:
`packages/old` clutters up searching through the monorepo. Wondering if we can get rid of it, now that `old` is ~2 months old at this point.


frankhinek:
@amika-sq have been wanting to do this for a while

its only there out of convenience, and I've got `packages/old/**/*,` in my VSCode "files to exclude" to minimize the annoyance when searching

The last thing I'm pulling out old that needs to go back into the new packages is the Connect Protocol code (original implementation was from 6+ months ago, pre polyrepo).

but safe to remove since we can always pull it from an earlier point in the commit history

mistermoe (approved):
💯

PR #240, opened 2023-10-06
Add DID DHT method

decentralgabe:
https://github.com/TBD54566975/did-dht-method


decentralgabe:
seeing ```error TS2209: The project root is ambiguous, but is required to resolve export map entry '.' in file '/home/runner/work/web5-js/web5-js/packages/dids/package.json'. Supply the `rootDir` compiler option to disambiguate.```

🤔

package.json
```diff
@@ -24,5 +24,10 @@
   "license": "Apache-2.0",
   "devDependencies": {
     "@npmcli/package-json": "5.0.0"
+  },
+  "dependencies": {
```
decentralgabe, line 28: `npx depcheck` complained about these, but can remove them

decentralgabe:
updated to use dns packets as per https://hackmd.io/-fBsfngHRpuJilwf5GY3Qg?both#DIDs-as-a-DNS-Packet

cc: @csuwildcat @Nuhvi

packages/dids/src/dht.ts
```diff
@@ -0,0 +1,339 @@
+import {Jose, PublicKeyJwk, Web5Crypto} from '@web5/crypto';
+import DHT from 'bittorrent-dht';
+import ed from 'bittorrent-dht-sodium';
+import {DidDocument} from './types.js';
```
mistermoe, line 4:
```suggestion
import type {DidDocument} from './types.js';
```

mistermoe:
@decentralgabe convention for imports in this repo is:

1. external dep types
2. external deps
3. internal dep types
4. internal deps

[example](https://github.com/TBD54566975/web5-js/blob/main/packages/dids/src/did-ion.ts#L1-L12)

anything i'm missing w.r.t. to imports @frankhinek ?

packages/dids/.mocharc.json
```diff
@@ -1,5 +1,9 @@
...
   "enable-source-maps": true,
   "exit": true,
-  "spec": ["tests/compiled/**/*.spec.js"]
-}
\ No newline at end of file
+  "spec": ["tests/compiled/**/*.spec.js"],
+  "node-option": [
+    "experimental-specifier-resolution=node",
+    "loader=ts-node/esm"
+  ]
```
mistermoe, line 8: any specific reason for using `ts-node`? we've gotten by without having to use it so far.

packages/dids/src/dht.ts
```diff
@@ -0,0 +1,339 @@
+import {Jose, PublicKeyJwk, Web5Crypto} from '@web5/crypto';
```
mistermoe, line 1: I think `PublicKeyJwk` and `Web5Crypto` are types which we can import explicitly as types.

packages/dids/.mocharc.json
decentralgabe: it lets me run mocha tests in webstorm really simply

happy to remove it and just re-add locally / remove before i push if easier

packages/dids/src/did-dht.ts
```diff
@@ -0,0 +1,225 @@
+import z32 from 'z32';
+import crypto from 'crypto';
```
mistermoe, line 2: `crypto` isn't available in the browser unless we polyfill it. AFAICT you're only using `crypto` to generate digests which we're using [`@noble/hashes`](https://www.npmjs.com/package/@noble/hashes) for. [examples here](https://github.com/search?q=repo%3ATBD54566975%2Fweb5-js+%40noble%2Fhashes%2Fsha&type=code)

@frankhinek, i forget, did we end up deciding it'd be worth adding common hashing functions to `common`?

packages/dids/src/did-dht.ts
```diff
@@ -0,0 +1,225 @@
...
+      }
+
+      case 'secp256k1': {
+        cryptoKeyPair = await new EcdsaAlgorithm().generateKey({
+          algorithm   : {name: 'ECDSA', namedCurve: 'secp256k1'},
+          extractable : true,
+          keyUsages   : ['sign', 'verify']
+        });
+        break;
+      }
```
mistermoe, line 163: nothing immediate just wanted to call out that the same code exists [here](https://github.com/TBD54566975/web5-js/blob/main/packages/dids/src/did-key.ts#L644-L666) for `did:key` and [here](https://github.com/TBD54566975/web5-js/blob/f3c7184bab45972821c89d6275eda8a6c295884f/packages/dids/src/did-ion.ts#L260-L282) for `did:ion`

@frankhinek thoughts on eventually adding a `Web5SubleCrypto` api surface that matches [W3C WebCrypto's SubtleCrypto](https://www.w3.org/TR/WebCryptoAPI/#subtlecrypto-interface)?

packages/dids/.mocharc.json
mistermoe: ah gotcha. did you have to configure something specific to get things to work with webstorm? i'm thinking we can get webstorm to continue working really simply without introducing `ts-node`. 

p.s. we should move the `ts-node` dependency to `devDependencies` in `package.json`

packages/dids/src/did-dht.ts
decentralgabe: worked for me

packages/dids/src/dht.ts
decentralgabe: cool didn't know about type imports, what's it do?

packages/dids/.mocharc.json
decentralgabe: yeah had to manually add a run config, happy to work with you to find another solution

mistermoe:
kk @decentralgabe it looks like `bittorrent-dht` and `bittorrent-dht-sodium` are not browser compatible. 

`bittorrent-dht` relies on:
* `crypto`
* `dgram`
* `fs`
* `net`
* `dns`
* `path`
* `os`

if `bittorrent-dht` is a client used to send UDP requests, we won't be able to polyfill our way through getting it to work anyway because browsers can't make UDP requests. we'd need to go through an http gateway

`bittorent-dht-sodium` relies on:
* `node-gyp`

 it's using `node-gyp` to wrap `libsodium` and surface it to nodejs. we have the necessary primitives needed to hand roll this  given all of the work @frankhinek has done on the Ed25519 front. 

from what i can tell (could be wrong)`bittorrent-dht-sodium` isn't [doing much](https://github.com/emilbayes/bittorrent-dht-sodium/blob/master/index.js) aside from generating Ed25519 keypairs and providing sign/verify functions (all of which we have). what may be tricky is supporting this:

```js
const dht = new DHT({ verify: eddsa.verify })
```

given that `eddsa.verify` is coming out of `bittorrent-dht-sodium`. will have to go look at the function signature of `eddsa.verify` to see if we can provider our own `verify` function. really hoping its `Uint8Array` in / `Uint8Array` out

cc: @frankhinek @csuwildcat

mistermoe:
p.s. CI failing now bc of linting issues in `dht.js`. should be able to configure webstorm to autofix on save which will take care of the spacing issues

csuwildcat:
@decentralgabe we definitely need to get this browser and React Native compatible before pulling it in.

package.json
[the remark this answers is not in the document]
decentralgabe: thanks

decentralgabe:
without `bittorrent-dht` we lose the ability to publish/read all together. if this is necessary we'll need to rely on our own gateways which surface a traditional HTTP API over using the DHT directly, which is not ideal.

decentralgabe:
updated to use the `pkarr` lib and remove reliance on `bittorrent-dht` direclty

packages/dids/build/buffer-polyfill.cjs
```diff
@@ -0,0 +1,3 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
decentralgabe, line 1: @kirahsapong @mistermoe did this for the pollyfill and it works great

decentralgabe:
ready to go pending final review/approval by @mistermoe @frankhinek

mistermoe:
@frankhinek any reservations on pulling this in?

frankhinek:
Just started review.  Also asked @shamilovtim to take a look to see what might need to be done to ensure that the `@web5/dids` package is still usable under React Native after this is merged.

shamilovtim:
I'll be taking a peak at this ASAP 👍

related: https://github.com/TBD54566975/web5-js/pull/240

packages/dids/package.json
```diff
@@ -76,41 +76,36 @@
   "dependencies": {
     "@decentralized-identity/ion-pow-sdk": "1.0.17",
     "@decentralized-identity/ion-sdk": "1.0.1",
+    "@noble/hashes": "^1.3.2",
+    "bittorrent-dht": "^11.0.5",
```
frankhinek, line 80: @decentralgabe any particular reason to list `bittorrent-dht` as an explicit dependency?  It appears to be a transitive dependency of the `pkarr` package.  Could lead to version conflicts to have it specified in `@web5/dids`.

packages/dids/package.json
```diff
@@ -76,41 +76,36 @@
...
+    "bittorrent-dht-sodium": "^1.2.0",
```
frankhinek, line 81: @decentralgabe any particular reason to list `bittorrent-dht-sodium` as an explicit dependency?  It doesn't appear to be used anywhere in the PR nor is it a dependency of any other packages as best I can tell.

packages/dids/package.json
```diff
@@ -76,41 +76,36 @@
...
-    "karma-firefox-launcher": "2.1.2",
-    "karma-mocha": "2.0.1",
-    "karma-mocha-reporter": "2.2.5",
-    "karma-webkit-launcher": "2.1.0",
     "mocha": "10.2.0",
-    "playwright": "1.36.2",
     "rimraf": "4.4.0",
     "sinon": "15.0.2",
-    "source-map-loader": "4.0.1",
     "typescript": "5.1.6"
```
frankhinek, line 109: @decentralgabe why were all of these dev dependencies removed?

packages/dids/src/dht.ts
```diff
@@ -0,0 +1,312 @@
...
+     * @param did The DID Document to publish
+     */
+  public static async publishDidDocument(keypair: Web5Crypto.CryptoKeyPair, did: DidDocument): Promise<boolean> {
+    const packet = await DidDht.toDnsPacket(did);
+    const pkarrKeypair = {
+      publicKey : keypair.publicKey.material,
+      secretKey : new Uint8Array([...keypair.privateKey.material, ...keypair.publicKey.material])
+    };
+    const signedPacket = SignedPacket.fromPacket(pkarrKeypair, packet);
+    const results = await Promise.all(PKARR_RELAYS.map(relay => Pkarr.relayPut(relay, signedPacket)));
```
frankhinek, line 25: @decentralgabe should the pkarr relay used to publish be a configurable parameter?  Fine with addressing this in a future PR.

packages/dids/src/dht.ts
```diff
@@ -0,0 +1,312 @@
...
+  }
+
+  /**
+     * Retrieves a DID Document from the DHT
+     * @param did The DID of the document to retrieve
+     */
+  public static async getDidDocument(did: string): Promise<DidDocument> {
+    const didFragment = did.replace('did:dht:', '');
+    const publicKeyBytes = new Uint8Array(z32.decode(didFragment));
+    for (const relay of PKARR_RELAYS) {
```
frankhinek, line 175: @decentralgabe should the pkarr relay used to retrieve the DID doc be a configurable parameter?  Fine with addressing this in a future PR.

frankhinek (changes requested):

packages/dids/package.json
decentralgabe: I ran `npx depcheck` and it told me to remove them - can revert if you want

packages/dids/package.json
decentralgabe: remnant, removing

packages/dids/package.json
decentralgabe: remnant, removing

packages/dids/package.json
```diff
@@ -76,41 +76,35 @@
   "dependencies": {
     "@decentralized-identity/ion-pow-sdk": "1.0.17",
     "@decentralized-identity/ion-sdk": "1.0.1",
+    "@noble/hashes": "^1.3.2",
```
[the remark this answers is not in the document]
decentralgabe: remnant, removing

packages/dids/src/dht.ts
decentralgabe: I believe it's something the library should take a stance on, but may be able to override

packages/dids/src/dht.ts
decentralgabe: same comment as above

decentralgabe: updated to support a configurable relay with a default set by the lib

packages/dids/src/dht.ts
decentralgabe: updated to support a configurable relay with a default set by the lib

frankhinek (approved):

PR #243, opened 2023-10-11
Implement VC creation in web5/api

ianpatton:
A minimal implementation for #183 and #184


ianpatton:
I looked at the other implementation in #148 but as far as I can tell, this can be done fairly simply if the create method just returns a JWT directly without worrying about storing it on the DWN. @web5/api already exposes functionality for storing records. 

From what I can tell, `processVcRequest` in the various agents exists mostly to deal with DWN. If storage were left to the user then maybe that can be removed from all agents?

PR #244, opened 2023-10-17
`Secp256k1` - disable malleability check when verifying

mistermoe:
relevant context here: https://github.com/TBD54566975/web5-kt/pull/87

the best course of action may be to provide callers with the option to enable/disable the malleability check. Just didn't have the time to plumb this optionality down through the crypto algorithms interface. Wanted to get this PR out so that Verifiable Credential signature verification doesn't 💩 out for creds signed in environments that don't guarantee a `low-s` signature.

cc: @andresuribe87


mistermoe:
@frankhinek can i get a stamp here?

diehuxx (approved):
thanks for the very detailed write up in the linked web5-kt issue.

frankhinek (approved):
LGTM

PR #252, opened 2023-11-01
Expose DWN PermissionsGrant Message

diehuxx:
Minimum permissions interface to expose to unblock delegated permissions grants.

`dwn.permissions.send()` will be extended in a future PR to to send other `Permissions` interface messages.

### DIFF
--- a/packages/agent/src/dwn-manager.ts
+++ b/packages/agent/src/dwn-manager.ts
@@ -6,6 +6,7 @@ import {
   RecordsWriteMessage,
   RecordsWriteOptions,
   Signer,
+  PermissionsGrant,
 } from '@tbd54566975/dwn-sdk-js';
 
 import { Jose } from '@web5/crypto';
@@ -69,6 +70,9 @@ const dwnMessageCreators = {
   [DwnInterfaceName.Records + DwnMethodName.Query]       : RecordsQuery,
   [DwnInterfaceName.Records + DwnMethodName.Write]       : RecordsWrite,
   [DwnInterfaceName.Records + DwnMethodName.Delete]      : RecordsDelete,
+  [DwnInterfaceName.Permissions + DwnMethodName.Grant]   : PermissionsGrant,
+  [DwnInterfaceName.Permissions + DwnMethodName.Request] : PermissionsGrant,
+  [DwnInterfaceName.Permissions + DwnMethodName.Revoke]  : PermissionsGrant,
   [DwnInterfaceName.Protocols + DwnMethodName.Query]     : ProtocolsQuery,
   [DwnInterfaceName.Protocols + DwnMethodName.Configure] : ProtocolsConfigure,
 };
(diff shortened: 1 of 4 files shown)


frankhinek:
@diehuxx Reviewed the PR and had one question that I was hoping to get your insight on:

The approach to sending a `PermissionsGrant` is different than how both Record and ProtocolsConfigure messages are sent to a remote DWN.  What was the motivation for introducing a new `message` property to the send request object?

packages/api/src/dwn-api.ts
```diff
@@ -91,6 +106,11 @@ export type RecordsWriteRequest = {
   store?: boolean;
 }
 
+export type PermissionGrantRequest = {
```
frankhinek, line 109: @diehuxx This type appears to be a duplicate.  It isn't used anywhere as best I can tell.  Should it be removed?

packages/api/src/dwn-api.ts
```diff
@@ -19,6 +20,20 @@ import { DwnInterfaceName, DwnMethodName } from '@tbd54566975/dwn-sdk-js';
 import { Record } from './record.js';
 import { Protocol } from './protocol.js';
 import { dataToBlob } from './utils.js';
+import { PermissionsGrant } from '@tbd54566975/dwn-sdk-js';
+import { PermissionsGrantMessage } from '@tbd54566975/dwn-sdk-js';
+import { PermissionsGrantOptions } from '@tbd54566975/dwn-sdk-js';
```
frankhinek, line 25: Can these be consolidated into the existing import statement a few lines up?

packages/agent/src/dwn-manager.ts
```diff
@@ -184,8 +188,11 @@ export class DwnManager {
       });
       dwnRpcRequest.message = message;
       messageData = data;
-
+    } else if ('message' in request) {
```
frankhinek, line 191: @diehuxx The approach to sending a `PermissionsGrant` is different than how both Record and ProtocolsConfigure messages are sent to a remote DWN. What was the motivation for introducing a new message property to the send request object?

packages/api/tests/dwn-api.spec.ts
```diff
@@ -569,4 +571,57 @@ describe('DwnApi', () => {
       });
     });
   });
+
+  describe('permissions.grant()', () => {
```
frankhinek, line 575: Could a test for sending a `PermissionsGrant` to a remote DWN be added in this PR?

packages/agent/src/dwn-manager.ts
diehuxx: I've been wanting to make that addition separate of `PermissionsGrants`. The existing code forces an existing message to be turned into `options` then re built and signed. Why not just send the existing message?

packages/api/tests/dwn-api.spec.ts
diehuxx: If I understand agents correctly, setting `dwn.permissions.grant(target  : bobDid.did, ...)` will send to a remote. We do that in the test in `describe('target: did'`

PR #255, opened 2023-11-02
Refactoring credentials api interface

nitro-neal:
This change brings the web5-js credentials package more in line with the web5-kt credentials package api interface.

This simplifies the creation of VCs and moves signing into a separate function call:

```javascript
const vc = VerifiableCredential.create("ExampleCredential", "did:issuer", "did:subject", myData)
const signedVc = vc.sign(signOptions)
```

This also moves Presentation Exchange into its own object to do anything with Presentation Exchange:

```javascript
const presentationResult = PresentationExchange.createPresentationFromCredentials(vcJwts, presentationDefinition)
```


frankhinek:
Pending feedback from @andresuribe87 / @mistermoe / @nitro-neal

frankhinek (approved):
LGTM  :shipit:

PR #258, opened 2023-11-03
did dht mock publishing

decentralgabe:
mock out time consuming and flaky network calls


mistermoe (approved):

frankhinek (approved):
LGTM :shipit:

Nice work @decentralgabe
- Node and Browser env tests both pass locally
- `test:node` runtime: 5 seconds
- `test:browser` runtime: 12 seconds

PR #262, opened 2023-11-06
Add PBKDF2 to `@web5/crypto`

frankhinek:
## PR Description

As noted by @leordev in PR #224 the [`@noble/hashes`](https://github.com/paulmillr/noble-hashes) implementation of PBKDF2 runs significantly slower than the native WebCrypto API and Node.js implementations.

I compared the average performance of the following implementations of PBKDF2:
-  `@noble/hashes` in both Node.js 18 and Chrome web browser
- WebCrypto API (aka "Subtle crypto") in the Chrome web browser
- `node:crypto` in Node.js 18

Screenshot 2023-11-06 at 8 44 28 AM

The performance difference is so drastic that I had to use log scale to even be able to see the `WebCrypto` and `node:crypto` values.  Even at a relatively modest (and unsafe) number of 100,000 iterations the `@noble` performance is 28 to 71 times slower.

At 500,000 iterations the `@noble` implementation is unusable, taking 3 to nearly 8 full seconds in web browser and Node.js, respectively.  Additionally, both @leordev and @csuwildcat have reported that it causes the browser to freeze and pause rendering.

Based on the great work done by @leordev in PR #224, this PR moves the PBKDF2 functionality into the `@web5/crypto` package, which enables it to be used both standalone and in a `CryptoManager` implementation.

### DIFF
--- a/packages/crypto/src/crypto-primitives/pbkdf2.ts
+++ b/packages/crypto/src/crypto-primitives/pbkdf2.ts
@@ -1,0 +1,80 @@
+import { crypto } from '@noble/hashes/crypto';
+
+import { isWebCryptoSupported } from '../utils.js';
+
+type DeriveKeyOptions = {
+  hash: 'SHA-256' | 'SHA-384' | 'SHA-512',
+  password: Uint8Array,
+  salt: Uint8Array,
+  iterations: number,
+  length: number
+};
+
+export class Pbkdf2 {
+  public static async deriveKey(options: DeriveKeyOptions): Promise<Uint8Array> {
+    if (isWebCryptoSupported()) {
+      return Pbkdf2.deriveKeyWithWebCrypto(options);
+    } else {
+      return Pbkdf2.deriveKeyWithNodeCrypto(options);
+    }
+  }
+
+  private static async deriveKeyWithNodeCrypto(options: DeriveKeyOptions): Promise<Uint8Array> {
+    const { password, salt, iterations } = options;
+
+    // Map the hash string to the node:crypto equivalent.
+    const hashToNodeCryptoMap = {
+      'SHA-256' : 'sha256',
+      'SHA-384' : 'sha384',
+      'SHA-512' : 'sha512'
+    };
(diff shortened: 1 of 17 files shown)


packages/crypto/tests/crypto-primitives.spec.ts
```diff
@@ -525,6 +527,125 @@ describe('Cryptographic Primitive Implementations', () => {
     });
   });
 
+  describe.only('Pbkdf2', () => {
```
mistermoe, line 530:
```suggestion
  describe('Pbkdf2', () => {
```

packages/crypto/src/crypto-primitives/pbkdf2.ts
```diff
@@ -0,0 +1,78 @@
...
+      'SHA-384' : 'sha384',
+      'SHA-512' : 'sha512'
+    };
+    const hash = hashToNodeCryptoMap[options.hash];
+
+    // Convert length from bits to bytes.
+    const length = options.length / 8;
+
+    // Dynamically import the `crypto` package.
+    const { pbkdf2 } = await import('node:crypto');
```
frankhinek, line 37: @shamilovtim I tried swapping this for just `crypto` but it causes the browser bundle build to fail:

```
> rimraf dist/browser.mjs dist/browser.js && node build/bundles.js

✘ [ERROR] Could not resolve "crypto"

    src/crypto-primitives/pbkdf2.ts:37:36:
      37 │     const { pbkdf2 } = await import('crypto');
         ╵                                     ~~~~~~~~

  The package "crypto" wasn't found on the file system but is built into node. Are you trying to
  bundle for node? You can use "platform: 'node'" to do that, which will remove this error.
  ```

Any tricks you can think of to still polyfill with `react-native-quick-crypto` for RN?

shamilovtim: Let me try sending in a PR for this

shamilovtim:
RN is good to go

PR #264, opened 2023-11-06
avoid bundling crypto on browser with esbuild

shamilovtim:
In theory it should be simple to make a package isomorphic such that

```js
const crypto;

if (isNode)
  crypto = async import('crypto');
else
  crypto = async import('crypto').subtle;
```

However, when bundling with esbuild for our browser bundles it has the unfortunate side effect of evaluating conditional `import` behind an `if` that would typically never be evaluated during run time. In order to resolve this we can use the [external](https://esbuild.github.io/api/#external) directive from esbuild in order to never bundle crypto and evaluate it at compile time instead. I can't imagine we would ever want to bundle crypto at all regardless of isomorphism so it seems that adding this directive to our bundle configs should be a net positive.


shamilovtim:
Closing after discussion with @frankhinek we do in fact have a few limited places where we polyfill `crypto` for browser bundles. We can revisit another time when all `crypto` polyfills have been removed

PR #267, opened 2023-11-07
Re-introduce `resolve` Method in `DidApi` for DID Resolution

frankhinek:
## Summary
This PR (re-)introduces the `resolve()` method to the `DidApi` class, enabling the resolution of DIDs through `web5.dids.resolve()`. The method is designed to be a simple interface for resolving a DID to its corresponding DID Resolution Result, which includes the DID Document.

## Implementation Details
- The `DidApi` class is equipped with a `resolve` method that takes a DID URL and optional resolution options.
- The resolution process is carried out by delegating the request to the `Web5Agent`, which is expected to handle the actual resolution logic.
- The `resolve` method returns a promise that resolves with the `DidResolutionResult`.

### DIFF
--- a/packages/api/src/did-api.ts
+++ b/packages/api/src/did-api.ts
@@ -1,51 +1,14 @@
 import type { Web5Agent } from '@web5/agent';
+import type { DidResolutionOptions, DidResolutionResult } from '@web5/dids';
 
-// import type {
-//   DidKeyOptions,
-//   DidIonCreateOptions,
-//   DidMethodApi,
-//   DidMethodCreator,
-//   DidMethodResolver,
-//   DidResolverCache,
-//   DidResolutionResult,
-//   DidState
-// } from '@web5/dids';
-
-
-
-// import { DidResolver } from '@web5/dids';
-
-// // Map method names to option types
-// type CreateMethodOptions = {
-//   ion: DidIonCreateOptions;
-//   key: DidKeyOptions;
-// };
-
-// // A conditional type for inferring options based on the method name
-// type CreateOptions<M extends keyof CreateMethodOptions> = CreateMethodOptions[M];
-
-// export type DidApiOptions = {
-//   didMethodApis: DidMethodApi[];
-//   cache?: DidResolverCache;
-// }
+import { DidMessage } from '@web5/agent';
 
 /**
- * The DID API is used to create and resolve DIDs.
+ * The DID API is used to resolve DIDs.
  *
  * @beta
  */
 export class DidApi {
-  // private didResolver: DidResolver;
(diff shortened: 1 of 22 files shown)


packages/user-agent/src/user-agent.ts
```diff
@@ -213,7 +223,7 @@ export class Web5UserAgent implements Web5ManagedAgent {
     throw new Error('Not implemented');
   }
 
-  async sendDidRequest(_request: SendDidRequest): Promise<DidResponse> {
+  async sendDidRequest(_request: DidRequest): Promise<DidResponse> {
     throw new Error('Not implemented');
```
nitro-neal, line 227: how does this relate to 
rpc-client.sendDidRequest() ?

packages/api/src/did-api.ts
```diff
@@ -54,54 +17,20 @@ export class DidApi {
...
-  // addMethodCreator(_creator: DidMethodCreator) {
-  //   throw new Error('not yet implemented');
-  // }
+  /**
+   * Resolves a DID to a DID Resolution Result.
+   *
+   * @param didUrl - The DID or DID URL to resolve.
+   * @returns A promise that resolves to the DID Resolution Result.
+   */
+  async resolve(didUrl: string, resolutionOptions?: DidResolutionOptions): Promise<DidResolutionResult> {
```
nitro-neal, line 26: Will create be added eventually?

nitro-neal (approved):

nitro-neal (approved):
Looks good!

packages/api/src/did-api.ts
frankhinek: Yes, absolutely.  Main goal of this PR was to get the plumbing setup to restore usage of `web5.did.*` since Angie has requested for some time that it be re-introduced.

packages/user-agent/src/user-agent.ts
frankhinek: Its a good question and I think the use case for sending requests is different for DIDs than it is with DWNs.

With `web5.did.resolve()` there really isn't a scenario where you would send the DID resolution request to another remote agent.

Even once `web5.did.create()` is introduced, I can't imagine a scenario in which a `Web5UserAgent` would send a request to another agent asking it to create the DID (and keys).

So the only scenario would be something like a `Web5ProxyAgent` communicating with something like a desktop Identity Agent... In that case, in `packages/proxy-agent/src/proxy-agent.ts` you would expect that all of the calls to `processDidRequest` are actually calls to `Web5RpcClient.sendDidRequest()`, which then forwards the request over the network.

csuwildcat (approved):

PR #268, opened 2023-11-07
Support RecordsQuery pagination in Web5

LiranCohen:
Add `paginationMessageCid` to the `RecordsQueryResponse` object, now when providing a limit, if there are additional results this cursor will be returned to be used in subsequent queries.

Example response with additional results:
```javascript
{
  status: { code: 200, message: '...' },
  records: [...],
  paginationMessageCid: 'bayfreigu....'
}
```

Example request for additional results:
```javascript
const results = await dwn.records.query({
  message: {
    filter: {
      schema: 'foo/bar'
    },
    pagination: { limit: 2, messageCid: 'bayfreigu....' }
  }
});
```

Added a pagination test, as well as a missing sorts test.

@frankhinek I'm not sure if this is the best place to test these particular paths, input would be appreciated.

### DIFF
--- a/packages/api/src/dwn-api.ts
+++ b/packages/api/src/dwn-api.ts
@@ -133,7 +133,7 @@ export type RecordsQueryResponse = ResponseStatus & {
   records?: Record[]
 
   /** If there are additional results, the messageCid of the last record will be returned as a pagination cursor. */
-  paginationMessageCid?: string;
+  cursor?: string;
 };
 
 /**
@@ -363,7 +363,7 @@ export class DwnApi {
           agentResponse = await this.agent.processDwnRequest(agentRequest);
         }
 
-        const { reply: { entries, status, paginationMessageCid } } = agentResponse;
+        const { reply: { entries, status, cursor } } = agentResponse;
 
         const records = entries.map((entry: RecordsQueryReplyEntry) => {
           const recordOptions = {
@@ -384,7 +384,7 @@ export class DwnApi {
           return record;
         });
 
-        return { records, status, paginationMessageCid };
+        return { records, status, cursor };
       },
 
       /**
--- a/packages/api/src/dwn-api.ts
+++ b/packages/api/src/dwn-api.ts
@@ -132,7 +132,7 @@ export type RecordsQueryRequest = {
 export type RecordsQueryResponse = ResponseStatus & {
   records?: Record[]
 
-  /** If there are additional results, the messageCid of the last item will be returned as a pagination cursor. */
+  /** If there are additional results, the messageCid of the last record will be returned as a pagination cursor. */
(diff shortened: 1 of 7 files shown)


frankhinek (commented):
@LiranCohen Can you also update the README files as part of this PR once we settle on the `web5.dwn.*` API syntax?

- https://github.com/TBD54566975/web5-js/blob/main/README.md
- https://github.com/TBD54566975/web5-js/blob/main/packages/api/README.md

packages/api/src/dwn-api.ts
```diff
@@ -353,7 +356,7 @@ export class DwnApi {
           agentResponse = await this.agent.processDwnRequest(agentRequest);
         }
 
-        const { reply: { entries, status } } = agentResponse;
+        const { reply: { entries, status, paginationMessageCid } } = agentResponse;
```
frankhinek, line 359: @LiranCohen In the hopes of providing the most intuitive developer experience, I was thinking that perhaps we name `paginationMessageCid` something that alludes to this value being the pagination cursor.

I was reading through the PR that originally introduced the pagination feature.  I can understand the reasons for using the `messageCid` as the token but it seems like providing a more descriptive name to the value returned would have been more intuitive for developers using the feature.  Particularly since so many other pagination implementations often return to the value as a `token` or `cursor`.

Thoughts?

LiranCohen: @frankhinek 

I agree I think what we landed on wrt the cursor objects is less than optimal, especially looking at it again after some time.

I'm leaning towards a more generic naming, and we can define the explicitness of the values within docs/spec/comments.

Example:
```javascript
// Query pagination object in descriptor
export type Pagination = {
  cursor?: string
  limit?: number
};

export type RecordsQueryResponse = {
  status: UnionMessageReply['status'];
  records?: Record[];
  cursor?: string;
};
```

If we really want to be explicit about the cursor being a `messageCid`, which could have some value, maybe something like `cursorMessageCid`?

LiranCohen: @frankhinek Originally was thinking of making the change here first to get it in, but then realized that it would be a hassle since the pagination object with `messageCid` is in the descriptor of the request.

I opened a PR in `dwn-sdk-js` to get this in and bubble it up to here.

LiranCohen:
@frankhinek updated this PR and added the missing documentation. Although would love some suggestions on the wording, not my strongest suit.

Also, not sure that I saw a pattern anywhere to document `enum` values like the `DateSort` enum we expose for sorting. Any suggestions on where/how to call it out or is it ok to just call out the type name like I did?

frankhinek (approved):
LGTM 🚀

csuwildcat (approved):

PR #269, opened 2023-11-09
Adding web vc create

nitro-neal:
This PR adds the web5.vc.create api. This currently uses the agents connected did's key to generate a VC to the subject specified, and returns a signed vcJwt. In the future this will write as a DWN record.

## Parameters

```js
  /**
   * Issues a VC to the subject did
   *
   * @param issuer The issuer URI of the credential, as a [String].
   * @param subject The subject URI of the credential, as a [String].
   * @param dataType The type of the credential, as a [String].
   * @param data The credential data, as a generic type [T].
   * @param expirationDate The expiration date.
   * @return A VerifiableCredential JWT.
   */
```

## Usage
```js
  vc = new VcApi({ agent: testAgent.agent, connectedDid: identityDid });
  const vcJwt = await vc.create(identityDid, identityDid, 'ExampleDataType', {example: 'goodStuff'});
```  

In future PRs I want to have a write to dwn flag, when this is set to true the VcResponse object will be extended with a dwn records object and the vcJwt.

### DIFF
--- a/packages/agent/src/index.ts
+++ b/packages/agent/src/index.ts
@@ -14,5 +14,6 @@ export * from './store-managed-key.js';
 export * from './store-managed-identity.js';
 export * from './sync-manager.js';
 export * from './utils.js';
+export * from './vc-manager.js';
 
 export * from './test-managed-agent.js';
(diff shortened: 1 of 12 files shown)


thehenrytsai:

PR #292, opened 2023-11-13
retry browser tests if they fail

finn-tbd:


frankhinek:
@shamilovtim working on replacing Karma that will eliminate the need for the lengthy timeouts

PR #305, opened 2023-11-20
add did resolver in memory cache when none is declared

shamilovtim:
Currently the Did Resolver Cache defaults to NOOP when none is supplied. This has implications for performance. Instead, provide a basic unpersisted in memory cache for the duration of the web5-js session.


shamilovtim:
Depends on #306 to fix the weird typescript flakes

frankhinek:
Addressed in #445 
Documentation included in `@web5/identity-agent` README that explains how to use an in-memory cache or to substitute for one that works with React Native.
Future PR will move the in-memory DID resolver cache into `@web5/dids`

PR #307, opened 2023-11-21
deflake web tests

shamilovtim:
deflake web tests. bump some timeouts and add retry.

PR #308, opened 2023-11-21
speed up ci

shamilovtim:
## Reason for PR
- CI was running slow. checked the CI steps and saw that node_modules wasn't being cached.
- reinstalling npm globally wasn't necessary as v18 LTS ships a modern version of npm
- actions can be bumped to latest versions

## Solution

- cache node_modules. the cache invalidation of `actions/setup-node` says it supports npm workspaces and works as-is with symlinked projects.
- remove npm install -g npm CI step
- bump action versions to v4
- include #307

### DIFF
--- a/.github/workflows/tests-ci.yml
+++ b/.github/workflows/tests-ci.yml
@@ -8,7 +8,7 @@ on:
     branches:
       - main
 
-  # Allows you to run this workflow manually from the Actions tab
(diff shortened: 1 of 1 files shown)


shamilovtim:
This made zero difference. It was the web tests with the timeout / retry that were taking forever.

PR #309, opened 2023-11-21
fix syncmanager typedefs to use abstractlevel

shamilovtim:
## Reason for PR
When using typedefs with LevelDB they're meant to adhere to the abstract-level interfaces rather than directly using the typedefs from the `Level` package (which is a specific implementation of abstract level).

## How PR was implemented
Used the same approach as dwn-js-sdk by pulling from the `AbstractLevel` interface and correctly annotating its generics

### DIFF
--- a/packages/agent/src/sync-manager.ts
+++ b/packages/agent/src/sync-manager.ts
@@ -1,16 +1,14 @@
-import type { BatchOperation } from 'level';
+import { Level } from 'level';
+import type { AbstractLevel, AbstractBatchOperation } from 'abstract-level';
 import type {
   EventsGetReply,
   GenericMessage,
   MessagesGetReply,
   RecordsWriteMessage,
 } from '@tbd54566975/dwn-sdk-js';
-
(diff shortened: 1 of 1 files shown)


packages/agent/src/sync-manager.ts
mistermoe, line 1: i thiink the desired convention for imports from top to bottom in this repo (though i'm sure there's evidence of the contrary in a few places bc of time constraints) is:

* all `import type`s first
* then all `import`s

mistermoe (commented):
left 1 small comment. other than that, looks great

packages/agent/src/sync-manager.ts
shamilovtim: We can probably run a formatter that does this in the future. Although past experience has been that import sorting is always a fail

shamilovtim: done

csuwildcat (approved):

LiranCohen (approved):

PR #323, opened 2023-11-29
decouple tbdocs commenter

leordev:
Now we also have the tbdocs report in the GH Action summary! 
image

https://github.com/TBD54566975/web5-js/actions/runs/7039346479?pr=323#summary-19158131202

- Decouples the docs reporter from the CI tests
- Decouples the PR comments generation from the tbdocs activity because secrets are not exposed to forked PRs
  - we could either create a server app to watch the PR and open a comment from the server (same [approach taken by CodeCov](https://github.com/codecov/codecov-action/releases/tag/v1.0.6) reports which are passing in our PRs!)
  - or, per the github workflows security recommendations, use the `workflow_run` trigger to safely execute code from the main branch, access secrets and be able to create the report comment
    - [Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)
    - [workflow_run usage example](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#using-data-from-the-triggering-workflow)

PS: the tbdocs report comment in the PR will only start to show up when we merge this PR to main. Unfortunately this is the normal behavior for new `workflow_run` definitions.

Working evidence: 
- TBDocs PR with a comment: https://github.com/TBD54566975/tbdocs/pull/89#issuecomment-1832688597
- TBDocs Commenter workflow_run triggered action logs: https://github.com/TBD54566975/tbdocs/actions/runs/7039271133/job/19157923330#step:5:12

### DIFF
--- a/.github/workflows/docs-ci.yml
+++ b/.github/workflows/docs-ci.yml
@@ -1,0 +1,53 @@
+name: Docs Continuous Integration
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  tbdocs-reporter:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
+
+      - name: Set up Node.js
+        uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # v3.6.0
+        with:
+          node-version: 18
+          registry-url: https://registry.npmjs.org/
+
+      - name: Install latest npm
+        run: npm install -g npm@latest
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Build all workspace packages
+        run: npm run build
+
+      - name: TBDocs Reporter
+        id: tbdocs-reporter-protocol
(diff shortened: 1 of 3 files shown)


.github/workflows/docs-ci.yml
```diff
@@ -0,0 +1,53 @@
+name: Docs Continuous Integration
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
```
shamilovtim, line 9:
```suggestion
on:
  push:
    branches:
      - main
```

shamilovtim (commented):
Should this run on `push` only rather than on every commit of every PR?

```yaml
on:
  push:
    branches:
      - main
```

.github/workflows/docs-ci.yml
leordev: if you do that you lose the nice docs checks report on your pr (which is the whole point of this CI action...) 
this is like a linting for your docs... I dont think you want to skip it...
example:  https://github.com/TBD54566975/web5-js/pull/269#issuecomment-1802976855

shamilovtim: Cool!

frankhinek (approved):
Caught up with @leordev about getting the next step done after this which is to publish the API docs to gh-pages.  Hope to get that prioritized for next week.

shamilovtim (approved):

Issue #232, opened 2023-09-29
Replace Karma with another runner for browser tests


frankhinek:
## Rationale:
- [Karma](https://github.com/karma-runner/karma) was deprecated on Apr 27, 2023
- Post deprecation, the project is "[not accepting new features or general bug fixes.](https://github.com/karma-runner/karma#karma-is-deprecated-and-is-not-accepting-new-features-or-general-bug-fixes)".
- Web5 SDK JS browser tests frequently timeout, particularly with Webkit, and have required significantly increasing the timeout on many of the Mocha/Chai tests (5, 10, and even 30 seconds).  Despite this some tests still timeout.  This leads to aggregate CI runtimes to be lengthened considerably.
- Web5 SDK JS browser tests take 20+ minutes to complete under GitHub CI which results in slow feedback loops.

## Work to be done:
- First, propose a new approach/solution and get feedback from the project maintainers.  Once an approach is agreed upon, then proceed with modifying the SDK repos and submit a PR.  There are a number of other changes being contemplated with respect to tests, so that context should be incorporated in the plan to replace Karma.

michaelneale:
its probably a little out dated but jasmine was my pick back in the day. Looking now Jest seems more popular (if that counts for anything)

frankhinek:
Yea, Jest is more popular but it seems a lot of projects are getting away from Jest.  One of the reason some seem to be gravitating towards Vitest is that the API is fully Jest compatible.  And it also supports the expect style tests from Chai, which means we wouldn't have to rewrite many of our tests to migrate.

Also Moe hates Jest so we had steered away from it early on 😄

shamilovtim:
Did a deep dive and my conclusion is Jest is the way to go
[TBD Testing Migration.pdf](https://github.com/TBD54566975/web5-js/files/13462108/TBD.Testing.Migration.pdf)

shamilovtim:
Found a way to replace Karma with zero changes to the test suite: https://github.com/TBD54566975/web5-js/pull/316/files. Just a POC for Agent, and still not as good as using Jest. But very simple to do.

frankhinek:
Closing since Karma was replaced with Web Test Runner in #316

PR #324, opened 2023-12-02
add watch mode for browser testing

shamilovtim:
## Reason for PR

Watch mode is a convenient way to have live reload for testing but is currently missing from the existing test solutions in the project

## Implementation

Use the esbuild API and CLI arguments to integrate web test runner watch with esbuild watch. 

Esbuild watch uses incremental compilation for speed. Using esbuild watch only the relevant files should be re-compiled and web test runner watches these files with its native watcher. 

## Demo

https://github.com/TBD54566975/web5-js/assets/3495974/7146c266-8547-4cec-a378-6bd75eccf2b9

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -51,6 +51,14 @@ module.exports = {
     'no-trailing-spaces'                       : ['error'],
     '@typescript-eslint/no-non-null-assertion' : 'off',
     '@typescript-eslint/ban-ts-comment'        : 'off',
-    'mocha/no-exclusive-tests'                 : 'warn'
-  }
+    'mocha/no-exclusive-tests'                 : 'warn',
+  },
+  overrides: [
+    {
+      files : ['*.cjs'],
+      rules : {
+        '@typescript-eslint/no-var-requires': 'off'
+      }
+    }
+  ]
 };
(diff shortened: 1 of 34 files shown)


shamilovtim:
relates to #232

.eslintrc.cjs
```diff
@@ -51,6 +51,14 @@ module.exports = {
     'no-trailing-spaces'                       : ['error'],
     '@typescript-eslint/no-non-null-assertion' : 'off',
     '@typescript-eslint/ban-ts-comment'        : 'off',
-    'mocha/no-exclusive-tests'                 : 'warn'
-  }
+    'mocha/no-exclusive-tests'                 : 'warn',
+  },
+  overrides: [
```
thehenrytsai, line 56: I think there is a point to allow var requires only in the .cjs files. Let's revert this part if you agree.

shamilovtim: that's what this does: allow var requires in CJS files

thehenrytsai (commented):
Pretty cool and powerful, would love to merge this in.

Though no one would probably know about it thus no one would use it unless it is promoted and documented beyond the PR description here.

I'd suggest to add a section in CONTRIBUTING.md (just copy and paste what you have in PR description, would be even better if video demo is intact too) to promote this as a tool at contributor's disposal.

.eslintrc.cjs
thehenrytsai: Does it not also allow var requires in non-cjs files?

shamilovtim: the changes here: `if cjs file allow requires`

thehenrytsai: I am dumb, thanks for pointing out the obvious...

LiranCohen (approved):
This is really awesome!

Same comments as @thehenrytsai about maybe just adding a blurb about it to CONTRIBUTING.md

PR #326, opened 2023-12-04
Changes in Readme

AnuragNishad12:


frankhinek:
Hi @AnuragNishad12.  Thanks for opening the PR.  This feature (`web5.did.create()`) is currently disabled.  We should likely remove it from the README until its added back and create an issue to track re-adding the text to the README.

frankhinek:
To be fixed in #445

PR #327, opened 2023-12-04
Fix Bugs with Reading Remote Records and Repeated Record Data Access

frankhinek:
## Summary

This PR primarily focuses on fixing several bugs reported by community members that occur when interacting with records retrieved from remote DWNs and subsequent attempts to consume the data payload.  The approach taken to resolve these issues additionally simplifies and improves the Web and Node Stream handling.

## Context

A number of community members and TBD developers had reported that errors were returned when attempting to update, delete, and read data from `Record` instances.  Further troubleshooting revealed that the root causes were related to records that had been initially retrieved from a remote DWN via `web5.dwn.records.query()` or that had been retrieved via `web5.dwn.records.read()`.  A secondary bug was introduced if the record exceeded the size limit that DWNs use to determine whether a `RecordsQuery` should included the data payloads Base64 URL encoded.

During the troubleshooting process, a tangentially related issue was raised that initially obscured the true root cause: repeated attempts to access a record's data via the `record.data.*` getter resulted in hangs/timeouts.  This was caused by the underlying `Readable` stream being consumed by the initial read and subsequent requests failing.

#### `@web5/common`
- Added a new `Stream` class to provide common functionality needed when working with [Web Streams API](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API) objects, most notably `ReadableStream`.  Implemented in an isomorphic manner that should work across Node.js, browser, and hopefully React Native runtime environments.
- Added a new `NodeStream` class to provide common functionality needed when working with [Node.js Stream](https://nodejs.org/dist/latest-v20.x/docs/api/stream.html) objects, most notably `Readable`.  Implemented in an isomorphic manner that should work across Node.js, browser, and hopefully React Native runtime environments.
- Improve test organization and coverage.

#### `@web5/api`
- Fixes a bug that occurs when a record that was initially retrieved in the query results from a remote DWN is interacted with.  The `Record` class now keeps track of whether the record's source is a remote target so that any requests to fetch the data payload are executed on the correct DWN.
- Fixes a bug that would cause subsequent attempts to access a `Record`'s data payload via `record.data.*` to fail as a result of the underlying `Readable` stream having already been consumed.
- Completely refactored the internals of the `Record.data` getting to remove confusing and error-prone handling of data.  In particular, the conditional handling of async method calls and unresolved Promises was removed.
- Adopted use of the new `@web5/common` stream capabilities that were introduced as part of the resolution of the aforementioned issues.
- Adds the `bytes()` method to the `Record.data` getter.  This enables developers to easily access the data payload of a record as a byte array in `Uint8Array` format.
- Updates the `README` to include details about both the `blob()` and `bytes()` methods.
- Removes the `record.delete()` instance method from `Record` objects.  This method is the only one that is a duplicate of the identically named `web5.dwn.records.*` method and introduces unintuitive and error prone behavior when the record was initially retrieved from a remote DWN.  Another class of issues is then introduced depending on whether that record already exists locally.  In order to remove this "foot gun" and keep the DWN Records API provided by `@web5/api` free of duplicates that behavior differently depending on unintuitive context, the instance method was removed.
- Improve documentation of `Record` class methods.
- Improve test coverage for edge cases.

### DIFF
--- a/packages/api/src/record.ts
+++ b/packages/api/src/record.ts
@@ -203,7 +203,7 @@ export class Record implements RecordModel {
        * Returns the data of the current record as a `Blob`.
        *
        * @returns A promise that resolves to a Blob containing the record's data.
-       * @throws {Error} If the record data is not available or cannot be converted to a `Blob`.
+       * @throws If the record data is not available or cannot be converted to a `Blob`.
        *
        * @beta
        */
@@ -215,7 +215,7 @@ export class Record implements RecordModel {
        * Returns the data of the current record as a `Uint8Array`.
        *
        * @returns A Promise that resolves to a `Uint8Array` containing the record's data bytes.
-       * @throws {Error} If the record data is not available or cannot be converted to a byte array.
+       * @throws If the record data is not available or cannot be converted to a byte array.
        *
        * @beta
        */
@@ -227,7 +227,7 @@ export class Record implements RecordModel {
        * Parses the data of the current record as JSON and returns it as a JavaScript object.
        *
        * @returns A Promise that resolves to a JavaScript object parsed from the record's JSON data.
-       * @throws {Error} If the record data is not available, not in JSON format, or cannot be parsed.
+       * @throws If the record data is not available, not in JSON format, or cannot be parsed.
        *
        * @beta
        */
@@ -239,7 +239,7 @@ export class Record implements RecordModel {
        * Returns the data of the current record as a `string`.
        *
        * @returns A promise that resolves to a `string` containing the record's text data.
-       * @throws {Error} If the record data is not available or cannot be converted to text.
+       * @throws If the record data is not available or cannot be converted to text.
        *
        * @beta
        */
@@ -251,7 +251,7 @@ export class Record implements RecordModel {
        * Provides a `Readable` stream containing the record's data.
        *
        * @returns A promise that resolves to a Node.js `Readable` stream of the record's data.
-       * @throws {Error} If the record data is not available in-memory and cannot be fetched.
+       * @throws If the record data is not available in-memory and cannot be fetched.
        *
        * @beta
        */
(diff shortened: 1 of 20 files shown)


csuwildcat (approved):

packages/api/src/record.ts
```diff
@@ -151,26 +150,39 @@ export class Record implements RecordModel {
...
     this._encryption = options.encryption;
     this._recordId = options.recordId;
 
+    if (options.encodedData) {
+      // If `encodedData` is set, then it is expected that:
+      // type is Blob if the Record object was instantiated by dwn.records.create()/write().
+      // type is Base64 URL encoded string if the Record object was instantiated by dwn.records.query().
+      // If it is a string, we need to Base64 URL decode to bytes and instantiate a Blob.
+      this._encodedData = (typeof options.encodedData === 'string') ?
+        new Blob([Convert.base64Url(options.encodedData).toUint8Array()], { type: this.dataFormat }) :
```
shamilovtim, line 176: I've noticed that in DWN and affiliated libs we use `base64url` rather than `base64` (atob/btoa). Is there a good reason for this?

frankhinek: @shamilovtim Base64 URL encoding is the format used in a variety of other related standards (e.g., JWK, DIDs, VCs, etc.).

Chat GTP answer:
The key difference between base64url and regular base64 encoding lies in the characters used for encoding. In regular base64 encoding (as used by atob and btoa in JavaScript), the characters + and / are used. However, these characters can cause issues in URLs and filenames since they have special meanings in those contexts.

base64url encoding addresses this by replacing + with - and / with _. This slight modification makes it URL and filename safe. The padding character = is also typically omitted in base64url to further enhance URL-friendliness.

shamilovtim: That's where I'm confused. I assume the DWN data is never used in a URL as its intended for file / data transfer. So should we consider the disadvantages of using [pure JS implementations of base64](https://github.com/multiformats/js-multiformats/blob/master/src/bases/base.js#L291C30-L291C30) over browser native ([atob/btoa implemented in C++](https://chromium.googlesource.com/external/github.com/flutter/engine/+/demo_apk_23/sky/engine/core/frame/DOMWindowBase64.cpp)) implementations of base64?

PR #334, opened 2023-12-05
DID DHT Enhancements

frankhinek:


packages/dids/src/did-dht.ts
```diff
@@ -284,7 +278,28 @@ export class DidDhtMethod implements DidMethod {
...
+     */
+    try {
+      didDocument = await DidDht.getDidDocument({ did: parsedDid.did });
+    } catch (error: any) {
+      return {
+        '@context'            : 'https://w3id.org/did-resolution/v1',
+        didDocument           : null,
+        didDocumentMetadata   : {},
+        didResolutionMetadata : {
+          contentType  : 'application/did+ld+json',
```
decentralgabe, line 297: it is not LD! just `application/did+json` should suffice

https://www.w3.org/TR/did-core/#application-did-json

decentralgabe (approved):

packages/dids/src/did-dht.ts
frankhinek: Done ✅

csuwildcat (approved):

PR #336, opened 2023-12-06
Export PEX type directly

kirahsapong:
This PR exports the sphereon `PresentationDefinitionV2` type directly, instead of a type alias, to resolve a TS issue inferring the type in consuming packages, where `xyz` should be of type `PresentationDefinitionV2`:
> ```The inferred type of 'xyz' cannot be named without a reference to '.pnpm/@sphereon+pex-models@2.0.3/node_modules/@sphereon/pex-models'. This is likely not portable. A type annotation is necessary.```

See [here](https://github.com/TBD54566975/tbdex-js/pull/112#issuecomment-1841107693) for further discussion

### DIFF
--- a/packages/credentials/src/presentation-exchange.ts
+++ b/packages/credentials/src/presentation-exchange.ts
@@ -1,13 +1,14 @@
 import type { EvaluationResults, PresentationResult, SelectResults, Validated as PexValidated } from '@sphereon/pex';
 import { PEX } from '@sphereon/pex';
 
-import type { PresentationDefinitionV2 } from '@sphereon/pex-models';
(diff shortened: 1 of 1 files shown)


nitro-neal:
I dont understand why typescript is doing this.. 

Chad says we can also try this:

```
export interface PresentationDefinitionV2 extends PexPresDefV2 {
    // Add any additional properties or methods if needed, or leave it empty
}
```

nitro-neal (approved):
If possible can we try this first? As it will look -more- uniform for the rest of the code. 
```
export interface PresentationDefinitionV2 extends PexPresDefV2 { }
```

and in the future we may change all exports to look like this

frankhinek (approved):

PR #338, opened 2023-12-07
bump dwn-sdk-js which pins uint8arrays to fix bug

shamilovtim:
Uses a backported fix to `dwn-sdk-js v0.2.8` (in order to avoid upgrading to 0.2.9 which has breaking changes)


shamilovtim:
fixed upstream

PR #339, opened 2023-12-07
Add JWT and DID `dereference()`

mistermoe:
# What's Changed

## `@web5/dids`

Added `DidResolver.dereference` [here](https://github.com/TBD54566975/web5-js/blob/verify-jwt/packages/dids/src/did-resolver.ts#L114-L154)

resolves the DID contained in the DID URL to a DID document, and then extracts and returns the specific part of the document identified by the fragment in the DID URL

---

Added a `DidResource` type [here](https://github.com/TBD54566975/web5-js/blob/verify-jwt/packages/dids/src/types.ts#L93-L96) which represents all of the resources present in a DID Document e.g. services, verification methods etc.

---

Added a `isVerificationMethod` utility function [here](https://github.com/TBD54566975/web5-js/blob/verify-jwt/packages/dids/src/utils.ts#L126-L133) that acts as a type guard for `DidResource`

## `@web5/credentials`

Separated all JWT related functionality from `VerifiableCredential` into a `Jwt` class [here](https://github.com/TBD54566975/web5-js/blob/679e74c6ff0d3072c9bf35ed3125a44150f199be/packages/credentials/src/jwt.ts)

---

Handrolled [`Jwt.verify`](https://github.com/TBD54566975/web5-js/blob/679e74c6ff0d3072c9bf35ed3125a44150f199be/packages/credentials/src/jwt.ts#L163-L207) and dropped `did-jwt` as a dependency. `did-jwt` is a chonker and uses a number of deprecated third party libs. 

---

Refactored `VerifiableCredential` and `PresentationExchange` methods to consistently use options objects.

---

Refactored `VerifiableCredential` methods to consistently be async.

---

> [!IMPORTANT]
> Removed `VerifiedJwt` type as it's no longer being used anywhere. this type was previously exported which means this is a breaking change for any downstream consumers who were using this type

> [!WARNING]
> - `VerifiableCredential.create` is now an async method.
> - `VerifiableCredential.verify` & `VerifiableCredentials.parseJwt` method signatures changed to take options object as input.
> - `PresentationExchange` method signatures changed to take options object as input.

## `@web5/crypto`

Added `JwtHeaderParams` and `JwtPayload` types.

---

Dramatically improves performance of `randomUuid()` function:

Before:
```
┌─────────┬───────────────────────┬──────────────┬────────────────────┬───────────┬─────────┐
│ (index) │       Task Name       │   ops/sec    │ Average Time (ns)  │  Margin   │ Samples │
├─────────┼───────────────────────┼──────────────┼────────────────────┼───────────┼─────────┤
│    0    │       'uuidv4'        │ '6,015,626'  │ 166.2337190237164  │ '±2.49%'  │ 601563  │
│    1    │  'utils.randomUuid'   │  '567,477'   │ 1762.1839950610192 │ '±17.89%' │  56748  │
└─────────┴───────────────────────┴──────────────┴────────────────────┴───────────┴─────────┘
```

After:
```
┌─────────┬───────────────────────┬──────────────┬────────────────────┬──────────┬─────────┐
│ (index) │       Task Name       │   ops/sec    │ Average Time (ns)  │  Margin  │ Samples │
├─────────┼───────────────────────┼──────────────┼────────────────────┼──────────┼─────────┤
│    0    │       'uuidv4'        │ '6,254,233'  │ 159.89170818373506 │ '±2.22%' │ 625424  │
│    1    │  'utils.randomUuid'   │ '13,852,305' │  72.1901496571959  │ '±1.51%' │ 1385231 │
└─────────┴───────────────────────┴──────────────┴────────────────────┴──────────┴─────────┘
```


packages/dids/src/did-resolver.ts
```diff
@@ -99,9 +104,41 @@ export class DidResolver {
...
         resolutionOptions
       });
+
       await this.cache.set(parsedDid.did, resolutionResult);
 
       return resolutionResult;
     }
   }
+
+  async deference(params: DereferenceParams): Promise<DidResource> {
```
frankhinek, line 114: Hey @mistermoe -- haven't reviewed this PR in detail since its still in draft but noticed one small typo

```suggestion
  async dereference(params: DereferenceParams): Promise<DidResource> {
```

packages/credentials/src/compact-jwt.ts
```diff
@@ -0,0 +1,224 @@
...
+    const signatureBytes = Convert.base64Url(encodedJwt.signature).toUint8Array();
+
+    const algorithmId = `${decodedJwt.header.alg}:${publicKeyJwk['crv'] || ''}`;
+    if (!(algorithmId in CompactJwt.algorithms)) {
+      throw new Error(`Verification failed: ${algorithmId} not supported`);
+    }
+
+    const { signer, options } = CompactJwt.algorithms[algorithmId];
+
+    const isLegit = await signer.verify({
```
nitro-neal, line 156: 😆 
```suggestion
    const isAuthentic = await signer.verify({
```

packages/dids/src/did-resolver.ts
mistermoe: ah thank you! 1am mistakes

packages/credentials/src/compact-jwt.ts
mistermoe: `is100pPure` lol

packages/credentials/src/compact-jwt.ts
```diff
@@ -0,0 +1,229 @@
...
+  }
+
+  /**
+   * Decodes a JWT without verifying its signature.
+   * @param params - Parameters for JWT decoding, including the JWT string.
+   * @returns Decoded JWT parts, including header and payload.
+   * @example
+   * const decodedJwt = CompactJwt.decode({ compactJwt: myJwt });
+   */
+  static decode(params: DecodeJwtParams) {
```
mistermoe, line 186: thinking it might make more sense to name this `parse` instead of `decode`. thoughts?

nitro-neal: I think decode is fine:
we use parse as taking the jwt and turning it into a vc:
```
parseJwt(vcJwt: string): VerifiableCredential 
```

so this is actually decoding like jwt.io yea?

KendallWeihe: I like decode here

amika-sq: This seems to return `decoded` AND `encoded` parts of the JWK. In it's current form, `parse` seems to make more sense, because it's not just decoding.

KendallWeihe: What's the use case for returning the `encoded` pieces? Since they're in the input `params` we're just splitting apart and then re-returning

mistermoe: great q @KendallWeihe. both encoded and decoded jwt's are used in `verify` [here](https://github.com/TBD54566975/web5-js/blob/verify-jwt/packages/credentials/src/compact-jwt.ts#L136-L150)

KendallWeihe: Yeah okay, I'm on team parse now

KendallWeihe: I mean the alternative is to split this into two methods, `decode` and maybe `expand` (or maybe `CompactJwt.spread()`?) there is not good antonym to "compact" ... so yeah, I'm team parse lol

packages/credentials/src/compact-jwt.ts
```diff
@@ -0,0 +1,229 @@
...
+  signer  : new EdDsaAlgorithm(),
+  options : { name: 'EdDSA' },
+  alg     : 'EdDSA',
+  crv     : 'Ed25519'
+};
+
+/**
+ * Class for handling Compact JSON Web Tokens (JWTs).
+ * This class provides methods to create, verify, and decode JWTs using various cryptographic algorithms.
+ */
```
KendallWeihe, line 68: Is there a spec doc for compact JWTs we could link to here?

mistermoe: [done](https://github.com/TBD54566975/web5-js/blob/verify-jwt/packages/credentials/src/compact-jwt.ts#L69)

packages/credentials/src/compact-jwt.ts
```diff
@@ -0,0 +1,229 @@
...
+ * Parameters for decoding a JWT.
+ */
+export type DecodeJwtParams = {
+  compactJwt: string
+}
+
+/**
+ * Represents a verified JWT with signer information, header, and payload.
+ */
+export type VerifiedJwt = {
```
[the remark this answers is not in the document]
mistermoe: good catch! we aren't. removing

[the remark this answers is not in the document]
mistermoe: done https://github.com/TBD54566975/web5-js/commit/a99e05e65baa35fc96d168803622f9c886b0e81c

packages/credentials/src/compact-jwt.ts
mistermoe: updated to use `parse`

packages/credentials/src/verifiable-credential.ts
```diff
@@ -250,20 +187,7 @@ export class VerifiableCredential {
...
-    }
-
-    const verificationResponse = await verifyJWT(vcJwt, {
-      resolver: tbdResolver
-    });
-
-    if (!verificationResponse.verified) {
-      throw new Error('VC JWT could not be verified. Reason: ' + JSON.stringify(verificationResponse));
-    }
+    await CompactJwt.verify({ compactJwt: vcJwt });
```
nitro-neal, line 190: Note: this could be by our design, but this verify doesn't validate the payload.`validatePayload` could be thrown on the end of this if needed.
This could be a legit signed jwt but an invalid vcjwt.

Also here is the old verifyJwt lib - https://github.com/decentralized-identity/did-jwt/blob/master/src/JWT.ts#L486 it didn't really validate the payload either as far as I can tell but it did go into seeing if it was self issued and it looked at sub

mistermoe: sry my response was confusing. we can totally validate payload inside `verify`. i mean in addition to that. like incase they need the subject or issuer did they don't have to go dig it out. or if they need to do something with payload after validation

KendallWeihe (approved):

packages/credentials/src/verifiable-credential.ts
mistermoe: done! https://github.com/TBD54566975/web5-js/commit/483f5ee8ac5338423e4fef48791d165ddcabc4e8

packages/credentials/src/verifiable-credential.ts
```diff
@@ -249,21 +186,20 @@ export class VerifiableCredential {
...
+    const { signerDid, payload } = await CompactJwt.verify({ compactJwt: vcJwt });
+    const vc = payload['vc'] as VcDataModel;
+    if (!vc) {
+      throw new Error('vc property missing.');
     }
 
-    const verificationResponse = await verifyJWT(vcJwt, {
-      resolver: tbdResolver
-    });
+    validatePayload(payload['vc']);
```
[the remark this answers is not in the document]
mistermoe: ah yeah good call @kirahsapong !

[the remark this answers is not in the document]
mistermoe: done: https://github.com/TBD54566975/web5-js/commit/3cc312f613d25dbc854926621da6459c2e92819c

mistermoe:
@frankhinek is taking a look at some of the tests breaking in `@web5/agent` due to version discrepancies

frankhinek (approved):

PR #345, opened 2023-12-09
Fix record author bug

frankhinek:
## Summary

This PR primarily focuses on fixing a bug reported by community members [on Discord](https://discord.com/channels/937858703112155166/1181706066636058654) that occurred when interacting with records retrieved from remote DWNs and subsequent attempts to consume the data payload.

## Context

A number of community members and TBD developers had reported that errors were returned when attempting to read data from `Record` instances:

`Uncaught Error: DwnManager: Signing key not found for author: 'did:ion:...'`

Further troubleshooting revealed that when large (> `DwnConstant.maxDataSizeAllowedToBeEncoded`) records authored/signed by one party (Alice) were written to another party's DWN (Bob), the recipient (Bob) was unable to access the data payload with the aforementioned error being thrown. This bug only surfaces when accessing the data (`record.data.*`) of a record signed by a different entity  a `Record` instance's data, which requires fetching the data from a remote DWN. Since the large (> `DwnConstant.maxDataSizeAllowedToBeEncoded`) data was not returned with the query as `encodedData`, the `Record` instance's data is not available and must be fetched from the remote DWN using a `RecordsRead` message.

What made this bug particularly difficult to track down is that the bug only surfaces when keys used to sign the record are different than the keys used to fetch the record AND both sets of keys are unavailable to the test Agent used by the entity that is attempting to fetch the record. In all of the other existing tests, the same test agent is used to store the keys for all entities (e.g., "Alice", "Bob", etc.) so the bug never surfaced.

#### `@web5/api`
- Remove potentially confusing use of `author` and `target` in `Record` class that had been acting as a stand in for the `connectedDid`.  This double-meaning has already introduced several bugs, so this PR removes the usage of the overloaded term and instead refers directly to the `connectedDid`, where applicable.
- Improve Record class test structure/efficiency and add failing test for newly discovered bug.
- Rename `remoteTarget` to `remoteOrigin` to improve intuitiveness of semantic meaning.

#### `@web5/agent`
- Minor typo fix in error message.

### DIFF
--- a/packages/agent/src/dwn-manager.ts
+++ b/packages/agent/src/dwn-manager.ts
@@ -225,7 +225,7 @@ export class DwnManager {
         dwnReply = await this.agent.rpcClient.sendDwnRequest(dwnRpcRequest as DwnRpcRequest);
         break;
       } catch(error: unknown) {
-        const message = (error instanceof Error) ? error.message : 'Uknown error';
+        const message = (error instanceof Error) ? error.message : 'Unknown error';
         errorMessages.push({ url: dwnUrl, message });
       }
     }
--- a/packages/api/src/dwn-api.ts
+++ b/packages/api/src/dwn-api.ts
@@ -273,7 +273,6 @@ export class DwnApi {
         // Remove target from inherited properties since target is being explicitly defined in method parameters.
         delete inheritedProperties.target;
 
-
         // If `data` is being updated then `dataCid` and `dataSize` must not be present.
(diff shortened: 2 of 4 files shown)


csuwildcat (approved):

LiranCohen (commented):
Looks good! That last test case confused me a bit. Didn't look too much into it, but should we be able to send that?

LiranCohen (approved):

PR #352, opened 2023-12-14
Generate docs with TBDocs

leordev:
@frankhinek please let me know if you are happy with this @web5/api docs generation automation.

Preview image:
image

Artifact to be downloaded and tested locally: https://github.com/TBD54566975/web5-js/actions/runs/7215382638

### DIFF
--- a/.github/workflows/docs-publish.yml
+++ b/.github/workflows/docs-publish.yml
@@ -32,62 +32,66 @@ jobs:
         with:
           node-version: 18
           registry-url: https://registry.npmjs.org/
-          cache: 'npm'
+          cache: "npm"
 
       - name: Install dependencies
-        run: |
-          npm ci
-          npm i jsdoc
-          npm i clean-jsdoc-theme
+        run: npm ci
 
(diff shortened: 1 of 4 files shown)


.github/workflows/docs-publish.yml
```diff
@@ -2,10 +2,11 @@
 name: Build and Deploy Docs
 
 on:
-  # Runs on pushes targeting the default branch
-  # push:
-  #   branches:
-  #     - main
+  # TODO: uncomment before merging
```
leordev, line 5: automatic docs publishing when a new release is cut

angiejones:
what do you all think about including all of the web5 packages here, namely dids, credentials, and crypto? Or should they have their own docs?

leordev:
> what do you all think about including all of the web5 packages here, namely dids, credentials, and crypto? Or should they have their own docs?

We are planning exactly on that, but first we need to fix the doc errors!

leordev:
Closing in favor of #402

PR #325, opened 2023-12-02
speed up ci

shamilovtim:
## Reason for PR
CI runs especially on the `test-with-browsers` job were taking between 15 to 25 minutes. Some common optimizations were unimplemented and parallelism was not being used for tests despite GH for open source being free.

## Implementation
- bump action versions
- add node cache 
- add playwright cache 
- replace docker with native `node` (see: https://github.com/TBD54566975/dwn-server/pull/94)
- parallelize browser tests
- only do the minimum amount of work necessary for the browser workflow (e.g. building cjs isnt necessary there)

## Results
- CI runs are now consistently between **230% to 380% faster**
- Duration variance of ~10 minutes has almost completely been eliminated. Runs now finish in a fairly consistent time frame. 
- CI duration went from **15min to 25min** (frequently 20min+) to **6min to 8min**

PR #356, opened 2023-12-19
wrap up ci speedup

shamilovtim:
## Context
I was monitoring CI performance since #325 got merged and saw one last improvement I could have made which is an additional parallel run. In some cases group B slows down the entire CI run by minutes. By splitting group B, this speeds up CI duration. Furthermore, by putting agent on its own, all runs should perform faster, even when group B is not underperforming.

## Implementation
Add group C. CI duration should decrease by about a minute on average.

## Results
In some cases now down to 5 min CI runs. About 25% faster than current runs.
Screenshot 2023-12-20 at 12 25 12 AM


shamilovtim:
@frankhinek added `--workspace packages/crypto-aws-kms` to the matrix runner

frankhinek (approved):
LGTM

LiranCohen (approved):
🥇 Love this!!

This will be helpful when we add more robust testing for sync which may take longer.

Issue #271, opened 2023-11-09
Refactor Key Management in `@web5/crypto`, `@web5/dids`, and `@web5/credentials`


frankhinek:
## Design Guidelines
- The key manager interface must be passed as an argument to all public API methods that require key material. e.g.
    - DID Creation
    - Data Signing
- MVP should include an in-memory Key Manager implementation.
- Consumers of our SDKs should be able to provide their own KeyManager implementations if desired.

## Feature Proposal Tasks:
- [x] Propose API surface design prior to implementation. This can be done by creating a draft PR for a respective feature and providing prototypal code or proposing the design in the comments
- [x] Approval Required from a minimum of 2 people

## Implementation Tasks:
- [x] Produce/update API reference documentation for any/all methods added to the public API surface. documenting private API methods is optional
- [x] Produce/update relevant example usage
- [x] Test coverage must be provided for all public API methods.
- [ ] Tests that consume shared test vectors must all be passing

frankhinek:
# Proposal

## General Design Choices

- Use JSON web key (JWK) as the key format.
    - Conversion utilities provided to convert between JWK and (1) Web Crypto API's `CryptoKey` and (2) raw `Uint8Array` keys.
- Use IANA [JSON Object Signing and Encryption](https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms) (JOSE) naming conventions for algorithm, curves, etc. whenever they exist.  JOSE _style_ names used when a standard name is draft/unpublished.
- All I/O that interacts with private or secret keys is done via reference using a `KeyIdentifier` type, which is a `string` value.  Concrete implementations can use any string as the key identifier (e.g. JWK thumbprint, UUID generated by hosted KMS, etc.).
- Provide a base `crypto` interface definition that concrete implementations conform to.
- Any implementation of the `crypto` interface can be passed as an argument to all public API methods that involve key material (e.g., DID creation, VC signing, general data signing/verification, etc.).

## Base Interface

Define a `CryptoApi` interface that all concrete implementations implement/extend:

```typescript
export interface CryptoApi {

  // Key Generation Operations
  generateKey(options: Web5Crypto.GenerateKeyOptions): Promise<Web5Crypto.KeyIdentifier>;

  // Key Derivation Operations
  deriveBits(options: Web5Crypto.DeriveBitsOptions): Promise<Uint8Array>;
[...]

PR #360, opened 2023-12-28
Refactor Key Management in `@web5/crypto` and add `@web5/crypto-aws-kms`

frankhinek:
## Summary

This PR is largely focused on implementing the proposal in #271.  Refer to that issue for more context.

#### `@web5/crypto`
- Removal of the class inheritance based `algorithms-api` and child classes.
- New `CryptoApi` interface and modular types for `KeyGenerator`, `Signer`, `Cipher`, `Hasher`, etc. introduced as proposed in #271.
- `LocalKmsCrypto` added that provides an implementation of `CryptoApi` that uses a local key store (memory is the default but can be customized to use a persistent store).
- Algorithm support retained for AES-CTR, AES-GCM, ECDSA, EdDSA, and SHA-2.
- JOSE portion of the library restructured to make way for expanded support to come for JWT, JWS, and JWE.
- Bump version to `0.3.0` due to breaking changes and need to incrementally update dependent Web5 JS packages.
- Bump crypto dependencies:
  - `@noble/ciphers` from `0.4.0` to `0.4.1`
  - `@noble/curves` from `1.2.0` to `1.3.2`
  - `@noble/hashes` from 1.3.2` to `1.3.3`
- Remove `@types/ed2curve` development dependency since `ed2curve` was removed a while ago.
- Test coverage remains at 100% -- only appears to have dropped due to type only files in the `jose` directory, which will be covered once JWT, JWS, and JWE functionality is added.

#### `@web5/crypto-aws-kms`
- New package that provides an implementation of `CryptoApi` that uses [AWS KMS](https://docs.aws.amazon.com/kms/latest/developerguide/overview.html) for remote storage of keys and signing operations
- Features mirror those of the existing Web5 KT [`AwsKeyManager`](https://github.com/TBD54566975/web5-kt/blob/main/crypto/src/main/kotlin/web5/sdk/crypto/AwsKeyManager.kt) implementation:
    - Generating `ES256K` private keys secured by AWS KMS.
    - Signing data with ECDSA secp256k1 using AWS KMS.
    - Local verification of ECDSA signatures.
- Designed to be extensible so that support for additional key operations and algorithms can be added.

#### `@web5/api`, `@web5/crypto`, `@web5/dids`
- Bump deprecated `sinon` package from `15.0.2` to `16.1.3`

#### `@web5/agent`, `@web5/credentials`
- Corrected missing dev dependency: `sinon@16.1.3`

#### `@web5/*`
- Updated root `tsconfig.json` compiler options `lib` from `ES6` to `ES2022`.


mistermoe (approved):
💯 🏅

.github/workflows/tests-ci.yml
```diff
@@ -82,7 +82,7 @@ jobs:
       matrix:
         include:
           - group: "A"
-            packages: "--workspace packages/agent --workspace packages/dids "
+            packages: "--workspace packages/agent --workspace packages/dids"
```
nitro-neal, line 85: need to add the new package here?

packages/crypto/src/primitives/ed25519.ts
```diff
@@ -477,28 +516,25 @@ export class Ed25519 {
...
    *
    * @returns A Promise that resolves to a boolean indicating whether the key
    *          corresponds to a valid point on the Edwards curve.
    */
-  private static async validatePublicKey(options: {
-    key: Uint8Array
-  }): Promise<boolean> {
-    const { key } = options;
-
+  private static validatePublicKey({ publicKey }: {
```
nitro-neal, line 532: is this supposed to be public?

packages/crypto/src/primitives/ed25519.ts
```diff
@@ -286,55 +299,96 @@ export class Ed25519 {
    * - `d`: The private key component, base64url-encoded.
    * - `x`: The derived public key, base64url-encoded.
    *
-   * Example usage:
-   *
+   * @example
    * ```ts
    * const privateKey = await X25519.generateKey();
```
nitro-neal, line 304:
```suggestion
   * const privateKey = await Ed25519.generateKey();
```

packages/crypto/src/primitives/secp256k1.ts
```diff
@@ -663,30 +713,28 @@ export class Secp256k1 {
...
-   * @param options - The options for the key validation.
-   * @param options.key - The private key to validate, represented as a Uint8Array.
+   * @param params - The parameters for the key validation.
+   * @param params.privateKeyBytes - The private key to validate, represented as a Uint8Array.
    *
    * @returns A Promise that resolves to a boolean indicating whether the private key is valid.
    */
-  private static async validatePrivateKey(options: {
-    key: Uint8Array
+  private static async validatePrivateKey({ privateKeyBytes }: {
```
nitro-neal, line 728: is this supposed to be public?

packages/crypto/src/primitives/secp256k1.ts
```diff
@@ -696,28 +744,25 @@ export class Secp256k1 {
...
-   * @param options.key - The public key to validate, represented as a Uint8Array.
+   * @param params - The parameters for the key validation.
+   * @param params.publicKeyBytes - The public key to validate, represented as a Uint8Array.
    *
    * @returns A Promise that resolves to a boolean indicating the public key's validity on
    *          the secp256k1 curve.
    */
-  private static async validatePublicKey(options: {
-    key: Uint8Array
+  private static async validatePublicKey({ publicKeyBytes }: {
```
nitro-neal, line 760: is this supposed to be public?

packages/crypto/tests/fixtures/test-vectors/aes-ctr/decrypt.json
```diff
@@ -0,0 +1,35 @@
+{
+  "description": "AES-CTR decrypt test vectors",
```
nitro-neal, line 2: eventually we can add these to the sdk-development test vectors

nitro-neal (approved):

nitro-neal (approved):
Great job, few minor things. I'll look over it again in detail but overall looks good!

.github/workflows/tests-ci.yml
frankhinek: @nitro-neal Not for that set of tests.  The new `@web5/crypto-aws-kms` package isn't bundled for nor usable in web browsers.

packages/crypto/tests/fixtures/test-vectors/aes-ctr/decrypt.json
frankhinek: Yep!  Started using the new format so that its easy to pull them in to `sdk-development`.

packages/crypto/src/primitives/ed25519.ts
frankhinek: Good question.  It was initially but was marked private when working on the new `Signer`, `AsymmetricKeyGenerator`, etc. types and attempting to get these classes to "satisfy" those type definition constraints.  While it technically worked, it had the unfortunate side effect of making TSDocs disappear (longer discussion about anonymous classes).  Will mark as public for now and see how much usage these functions get.

packages/crypto/src/primitives/secp256k1.ts
frankhinek: Good question.  It was initially but was marked private when working on the new `Signer`, `AsymmetricKeyGenerator`, etc. types and attempting to get these classes to "satisfy" those type definition constraints.  While it technically worked, it had the unfortunate side effect of making TSDocs disappear (longer discussion about anonymous classes).  Will mark as public for now and see how much usage these functions get.

PR #366, opened 2024-01-05
Add logic to JWT.verify to validate JWT is not expired

kirahsapong:
Tackles the TODO added as part of [this PR](https://github.com/TBD54566975/tbdex-js/pull/126/files#diff-88db2970c884747728964bdcea4d76ce78e8bde892003d9f13a59d22c59bff26R329-R332) to ensure JWT.verify() throws an error if the JWT payload includes an [`exp`](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.4) property and that `exp` property is in the past.


nitro-neal (approved):
good stuff thanks for fixing this, checking with Moe if this is the dream for verify but it looks good to me

angiejones:
yay! will update the docs to let people know they don't have to do this themselves anymore

diehuxx (approved):

mistermoe (approved):
💯

mistermoe:
cc: @frankhinek

PR #370, opened 2024-01-09
Bump `dwn-sdk-js` to version `0.2.10`

LiranCohen:
This PR updates the `dwn-sdk-js` library to the version `0.2.10`

With this update come:
- allow writing of pruned initial `RecordsWrite`, removing the need for a `writePrunedRecord` method in the `dwn-manager`
- update union types for `UnionMessageReply` to properly represent a union type of all replies.
- updates to the `EventLog`, no longer tracking an external watermark, but rather an internal one.


packages/agent/src/sync-manager.ts
```diff
@@ -239,7 +238,7 @@ export class SyncManagerLevel implements SyncManager {
 
     for (let job of pushJobs) {
       const [key] = job;
-      const [did, dwnUrl, watermark, messageCid] = key.split('~');
+      const [did, dwnUrl, _, messageCid] = key.split('~');
```
frankhinek, line 241: @LiranCohen Is there a reason we're still using the watermark in the key?
In other words, why not delete it rather than assigning it to a value we don't use? (`_`)

packages/api/src/dwn-api.ts
```diff
@@ -2,10 +2,12 @@ import type { DwnResponse, Web5Agent } from '@web5/agent';
...
   RecordsQueryOptions,
+  RecordsQueryReply,
   RecordsWriteMessage,
   RecordsWriteOptions,
   RecordsDeleteOptions,
   ProtocolsQueryOptions,
+  ProtocolsQueryReply,
   RecordsQueryReplyEntry,
   ProtocolsConfigureMessage,
   ProtocolsConfigureOptions,
```
frankhinek, line 13: Super minor nit but imports are sorted by length in the monorepo:
```suggestion
  RecordsQueryReply,
  RecordsReadOptions,
  ProtocolsQueryReply,
  RecordsQueryOptions,
  RecordsWriteMessage,
  RecordsWriteOptions,
  RecordsDeleteOptions,
  ProtocolsQueryOptions,
  RecordsQueryReplyEntry,
  ProtocolsConfigureMessage,
  ProtocolsConfigureOptions,
```

packages/api/package.json
```diff
@@ -76,9 +76,9 @@
     "node": ">=18.0.0"
   },
   "dependencies": {
-    "@tbd54566975/dwn-sdk-js": "0.2.8",
     "@web5/agent": "0.2.5",
     "@web5/common": "0.2.2",
+    "@tbd54566975/dwn-sdk-js": "0.2.10",
```
frankhinek, line 81: Package.json sorts alphabetically
```suggestion
    "@tbd54566975/dwn-sdk-js": "0.2.10",
    "@web5/agent": "0.2.5",
    "@web5/common": "0.2.2",
```

packages/agent/tests/sync-manager.spec.ts
```diff
@@ -1,16 +1,20 @@
...
-import * as sinon from 'sinon';
-import { RecordsQueryReply, RecordsWriteMessage } from '@tbd54566975/dwn-sdk-js';
+import sinon from 'sinon';
+import chai, { expect } from 'chai';
+import chaiAsPromised from 'chai-as-promised';
 
 import type { ManagedIdentity } from '../src/identity-manager.js';
+import type { PortableDid } from '@web5/dids';
+
+import { RecordsQueryReply, RecordsQueryReplyEntry, RecordsWriteMessage } from '@tbd54566975/dwn-sdk-js';
```
frankhinek, line 9: Minor nit.  Web5 JS imports are sorted:
- external package type imports
- external package imports
- internal type imports
- internal imports

```suggestion
import type { PortableDid } from '@web5/dids';
import type { RecordsQueryReply, RecordsQueryReplyEntry, RecordsWriteMessage } from '@tbd54566975/dwn-sdk-js';

import sinon from 'sinon';
import chai, { expect } from 'chai';
import chaiAsPromised from 'chai-as-promised';

import type { ManagedIdentity } from '../src/identity-manager.js';
```

packages/agent/package.json
```diff
@@ -68,10 +68,11 @@
...
   "dependencies": {
-    "@tbd54566975/dwn-sdk-js": "0.2.8",
+    "@tbd54566975/dwn-sdk-js": "0.2.10",
     "@web5/common": "0.2.2",
     "@web5/crypto": "0.2.2",
     "@web5/dids": "0.2.4",
+    "ulidx": "2.1.0",
     "level": "8.0.0",
     "readable-stream": "4.4.2",
     "readable-web-to-node-stream": "3.0.2"
```
frankhinek, line 78: Sorting
```suggestion
    "level": "8.0.0",
    "readable-stream": "4.4.2",
    "readable-web-to-node-stream": "3.0.2",
    "ulidx": "2.1.0"
```

packages/agent/src/sync-manager.ts
LiranCohen: @frankhinek Yeah, so the events come in the order they should be inserted when doing a `EventsGet`. We then stick them in a LevelDB queue, to read here.  The queue is sorted by the key, so the watermark here is used to sort it in the order it should be applied.

LiranCohen:
Thanks @frankhinek! I think I've addressed the formatting issues.

frankhinek (approved):

csuwildcat (approved):

PR #372, opened 2024-01-11
fix test vector

nitro-neal:


frankhinek:
Would prefer to improve the regex in `sdk-development` as opposed to applying a non-obvious constraint on test names

PR #375, opened 2024-01-13
Update Ed25519 and ES256K Test Vectors and Ensure Low-S ECDSA Signatures

frankhinek:
This PR will:
- Migrate to using Web5 SDK Test Vectors for `Ed25519` and `Secp256k1` `sign()` and `verify()` methods.
- Ensure ES256K signatures are low-S form in `@web5/crypto` and `@web5/crypto-aws-kms`.

> [!NOTE]
> A future PR will switch to using git submodules for Web5 SDK Test Vectors.

### DIFF
--- a/packages/crypto/tests/primitives/ed25519.spec.ts
+++ b/packages/crypto/tests/primitives/ed25519.spec.ts
@@ -4,9 +4,9 @@ import chaiAsPromised from 'chai-as-promised';
 
 import type { Jwk, JwkParamsOkpPrivate } from '../../src/jose/jwk.js';
 
-import ed25519Sign from '../fixtures/test-vectors/ed25519/sign.json' assert { type: 'json' };
-import ed25519Verify from '../fixtures/test-vectors/ed25519/verify.json' assert { type: 'json' };
+import CryptoEd25519SignTestVector from '../../../../test-vectors/crypto_ed25519/sign.json' assert { type: 'json' };
 import ed25519ComputePublicKey from '../fixtures/test-vectors/ed25519/compute-public-key.json' assert { type: 'json' };
+import CryptoEd25519VerifyTestVector from '../../../../test-vectors/crypto_ed25519/verify.json' assert { type: 'json' };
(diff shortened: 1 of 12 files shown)


mistermoe (approved):

packages/crypto/tests/primitives/ed25519.spec.ts
```diff
@@ -315,17 +315,24 @@ describe('Ed25519', () => {
...
-    for (const vector of ed25519Sign.vectors) {
-      it(vector.description, async () => {
-        const signature = await Ed25519.sign({
-          key  : vector.input.key as Jwk,
-          data : Convert.hex(vector.input.data).toUint8Array()
-        });
-
-        const signatureHex = Convert.uint8Array(signature).toHex();
-        expect(signatureHex).to.deep.equal(vector.output);
+    describe('Web5TestVectorsCryptoEd25519', () => {
```
nitro-neal, line 318: all the other examples are 1 level deep:
```
describe('PresentationExchange', () -> describe('Web5TestVectorsPresentationExchange', () 
describe('Verifiable Credential Tests', () -> describe('Web5TestVectorsCredentials', ()
```

yours are 2 levels deep
```
describe('Ed25519', () -> describe('verify()', () -> describe('Web5TestVectorsCryptoEd25519', ()
```


I -think- it should work and get to the checkmark screen, but if it doesn't then this will probably be the reason.

I think we should put all vector test in one file eventually for better ease of use, but we can discuss that offline

nitro-neal (approved):

packages/crypto/tests/primitives/ed25519.spec.ts
finn-tbd: junit xml doesnt have a way to do more than one class as far as i know, so the mocha junit reporter thing will just smush it down, which may or may not pass the regex. Probably best to just keep doing it the way we've established works.

frankhinek: Let's give it a shot and see how it does.

I had the same thought but the reason I ended up nesting them is that, otherwise, it isn't clear by looking at the tests that those methods are being rigorously tested.

PR #385, opened 2024-01-22
Record.store(), Record.import(), and roles API support

csuwildcat:
First pass at PR

### DIFF
--- a/packages/agent/src/dwn-manager.ts
+++ b/packages/agent/src/dwn-manager.ts
@@ -75,14 +75,6 @@ const dwnMessageConstructors = {
   [DwnInterfaceName.Protocols + DwnMethodName.Configure] : ProtocolsConfigure,
 };
 
-(RecordsWrite.prototype as any).forceSet = function(prop: string, value: unknown){
-  this[prop] = value;
-};
-
-(RecordsWrite.prototype as any).forceGet = function(prop: string){
-  return this[prop];
-};
-
 export type DwnManagerOptions = {
   agent?: Web5ManagedAgent;
   dwn: Dwn;
--- a/packages/agent/src/dwn-manager.ts
+++ b/packages/agent/src/dwn-manager.ts
@@ -283,7 +283,7 @@ export class DwnManager {
     });
 
     if (dwnMessageConstructor === RecordsWrite){
-      if (request.import) {
+      if (request.signAsOwner) {
         await (dwnMessage as RecordsWrite).signAsOwner(dwnSigner);
       }
     }
--- a/packages/agent/src/dwn-manager.ts
+++ b/packages/agent/src/dwn-manager.ts
@@ -196,9 +196,7 @@ export class DwnManager {
       messageData = data;
 
     } else {
-      console.log('start constructDwnMessage');
       const { message } = await this.constructDwnMessage({ request });
-      console.log('passed constructDwnMessage');
       dwnRpcRequest.message = message;
       messageData = request.dataStream;
     }
@@ -257,8 +255,6 @@ export class DwnManager {
   }) {
     const { request } = options;
     const rawMessage = (request as any).rawMessage;
-
-    console.log('259', rawMessage);
 
     let readableStream: Readable | undefined;
 
--- a/packages/agent/src/dwn-manager.ts
+++ b/packages/agent/src/dwn-manager.ts
@@ -242,11 +242,10 @@ export class DwnManager {
   }
 
   private async constructDwnMessage(options: {
-    request : ProcessDwnRequest
+    request: ProcessDwnRequest
   }) {
     const { request } = options;
-    const rawMessage = (request as any).rawMessage;
-
+    const rawMessage = request.rawMessage as any;
     let readableStream: Readable | undefined;
 
     // TODO: Consider refactoring to move data transformations imposed by fetch() limitations to the HTTP transport-related methods.
@@ -289,7 +288,7 @@ export class DwnManager {
       }
     }
 
-    return { message: (dwnMessage as any).message, dataStream: readableStream };
+    return { message: dwnMessage.message, dataStream: readableStream };
   }
 
   private async getAuthorSigningKeyId(options: {
--- a/packages/agent/src/dwn-manager.ts
+++ b/packages/agent/src/dwn-manager.ts
@@ -63,15 +63,24 @@ type DwnMessage = {
   data?: Blob;
 }
 
-const dwnMessageCreators = {
+const recordsWriteType = DwnInterfaceName.Records + DwnMethodName.Write;
+const dwnMessageConstructors = {
   [DwnInterfaceName.Events + DwnMethodName.Get]          : EventsGet,
   [DwnInterfaceName.Messages + DwnMethodName.Get]        : MessagesGet,
   [DwnInterfaceName.Records + DwnMethodName.Read]        : RecordsRead,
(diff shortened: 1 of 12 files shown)


frankhinek:
@csuwildcat Node and one of the browser tests are failing.  Can you look into why?

packages/api/src/record.ts
```diff
@@ -66,6 +69,20 @@ export type RecordUpdateOptions = {
...
+    const val = obj[key];
+    if (val === undefined) {
+      delete obj[key];
+    } else if (typeof val === 'object') {
+      removeUndefinedProperties(val);
+    }
+  });
+  return obj;
+}
+
```
frankhinek, line 85: To avoid duplication use the function that [exists](https://github.com/TBD54566975/web5-js/blob/2bd93a85754b6f39208968ce2fc464bf943f15c2/packages/common/src/object.ts#L35) in `@web5/common`.

`import { removeUndefinedProperties } from '@web5/common';`

packages/api/tests/record.spec.ts
```diff
@@ -1345,7 +1465,7 @@ describe('Record', () => {
     });
 
     // TODO: Fix after changes are made to dwn-sdk-js to include the initial write in every query/read response.
```
frankhinek, line 1467: TODO should be removed now that this test is passing.
```suggestion
```

packages/api/tests/record.spec.ts
```diff
@@ -97,6 +97,127 @@ describe('Record', () => {
...
+    //   from    : bobDid.did,
+    //   message : {
+    //     filter: {
+    //       schema: 'http://email-protocol.xyz/schema/email'
+    //     }
+    //   }
+    // });
+    // expect(bobQueryResult.status.code).to.equal(200);
+    // expect(bobQueryResult.records).to.exist;
+    // expect(bobQueryResult.records!.length).to.equal(1);
```
frankhinek, line 218: Why is this included but commented out?

packages/agent/src/dwn-manager.ts
```diff
@@ -242,17 +242,18 @@ export class DwnManager {
   }
 
   private async constructDwnMessage(options: {
-    request: ProcessDwnRequest
+    request : ProcessDwnRequest
```
frankhinek, line 245: Unnecessary space added in this PR that should be removed.
```suggestion
    request: ProcessDwnRequest
```

packages/api/src/record.ts
```diff
@@ -77,16 +94,45 @@ export class Record implements RecordModel {
...
+  private _initialWrite: RecordOptions['initialWrite'];
+  private _initialWriteStored: boolean;
   private _recordId: string;
-
+  private _protocolRole: RecordOptions['protocolRole'];
   // Getters for immutable DWN Record properties.
 
+  static sendCache = new Map();
+  static sendCacheLimit = 100;
+  static setSendCache(recordId, target){
```
frankhinek, line 109: Please add some documentation that explains what this does.
```suggestion
  static setSendCache(recordId, target) {
```

packages/api/src/record.ts
```diff
@@ -295,6 +350,91 @@ export class Record implements RecordModel {
...
+
+    if (options.dataStream) {
+      request.dataStream = options.dataStream;
+    }
+
+    return this._agent.processDwnRequest(request);
+  }
+
+
+  async store(options: any = {}): Promise<any> {
```
frankhinek, line 381: Please add some documentation that explains what this does.

```suggestion
  async store(options: any = {}): Promise<any> {
```

packages/api/src/record.ts
```diff
@@ -295,6 +350,91 @@ export class Record implements RecordModel {
     return dataObj;
   }
 
+  private async _processMessage(options: any){
+
+    const { store = true } = options;
```
frankhinek, line 355: Please add some documentation that explains why another `processMessage()` private method was added.

packages/api/src/record.ts
```diff
@@ -304,14 +444,55 @@ export class Record implements RecordModel {
...
-      target         : target,
-      messageOptions : this.toJSON(),
-    });
+  async send(_options?: string | { [key: string]: any }): Promise<ResponseStatus> {
+
+    const initialWrite = this._initialWrite;
+    const options = !_options ? { target: this._connectedDid } : typeof _options === 'string' ? { target: _options } : _options;
+
+    if (!options.target){
+      options.target = this._connectedDid;
```
frankhinek, line 453: This is some ugly looking code 😄 

It seems like there should be a better way to accomplish this.

packages/api/src/record.ts
```diff
@@ -304,14 +444,55 @@ export class Record implements RecordModel {
    *
    * @beta
    */
```
frankhinek, line 446: Please update the explanation for what `send()` does because it isn't the same as before.

packages/api/src/record.ts
```diff
@@ -304,14 +444,55 @@ export class Record implements RecordModel {
...
+        recordId      : this._recordId,
+        descriptor    : this._descriptor,
+        attestation   : this._attestation,
+        authorization : this._authorization,
+        encryption    : this._encryption,
+      });
+    }
+    else {
+      latestState.messageOptions = this.toJSON();
+    }
```
frankhinek, line 493: Please add inline comments that explain what is going on here so its easier to maintain in the future.

packages/api/src/record.ts
csuwildcat: Many tests fail when I switch to using the web5/common packages version.

csuwildcat: I am guessing it is because it assumes it will be a typed object of the one specified in the arguments of that version?

packages/api/src/record.ts
csuwildcat: Added

packages/api/src/record.ts
csuwildcat: Liran changed it

packages/api/tests/record.spec.ts
csuwildcat: Removed old unused test.

packages/agent/src/dwn-manager.ts
csuwildcat: Fixed

packages/api/src/record.ts
csuwildcat: Done

packages/api/src/record.ts
csuwildcat: Done

packages/api/src/record.ts
csuwildcat: Added

LiranCohen:
@frankhinek Would love another run through and blessing from you.

There is more clean up to do, specifically I'd love to get rid of `toJSON()` completely in favor of the `rawMessage` getter.
I'd like to simplify the messageOptions as well for record creation and re-think some of this flow now that we have more context.

But I think this is in a good place to un-block some bugs. And do further cleanup in subsequent PRs.

Would also love feedback from the community and users if the `import()` and `store()` methods make sense from a dev-ex perspective.

frankhinek (approved):

frankhinek:
> Would also love feedback from the community and users if the `import()` and `store()` methods make sense from a dev-ex perspective.

This is the biggest open question I still have.  Something about `import()` still seems confusing but better to ship and get feedback than endless deliberate.  We can always change it.

LiranCohen (approved):

PR #393, opened 2024-01-31
Refactor `@web5/dids` to support first class Key Management

### LINKED ISSUE #263: Add implementation of `did:web` to `@web5/dids`
## Design Guidelines
- Support for resolution of `did:web` DIDs by `DidWebMethod` implementation.
- Test that `DidWebMethod` can be used with the existing `DidResolver` implementation.

## Feature Proposal Tasks:
- [ ] Propose API surface design prior to implementation. This can be done by creating a draft PR for a respective feature and providing prototypal code or proposing the design in the comments
- [ ] Approval Required from a minimum of 2 people

## Implementation Tasks:
- [ ] Produce/update API reference documentation for any/all methods added to the public API surface. documenting private API methods is optional
- [ ] Produce/update relevant example usage
- [ ] Test coverage must be provided for all public API methods.
- [ ] Tests that consume shared test vectors must all be passing

### LINKED ISSUE #274: Add implementation of `did:jwk` to `@web5/dids`
## Design Guidelines
- Support for creation of `did:jwk` DIDs.
- Support for resolution of `did:jwk` DIDs by `DidJwkMethod` implementation.
- Test that `DidJwkMethod` can be used with the existing `DidResolver` implementation.

## Feature Proposal Tasks:
- [ ] Propose API surface design prior to implementation. This can be done by creating a draft PR for a respective feature and providing prototypal code or proposing the design in the comments
- [ ] Approval Required from a minimum of 2 people

## Implementation Tasks:
- [ ] Produce/update API reference documentation for any/all methods added to the public API surface. documenting private API methods is optional
- [ ] Produce/update relevant example usage
- [ ] Test coverage must be provided for all public API methods.
- [ ] Tests that consume shared test vectors must all be passing

### LINKED ISSUE #329: change `PortableDid`'s `did` property to `uri`
[`PortableDid`](https://github.com/TBD54566975/web5-js/blob/main/packages/dids/src/types.ts#L186-L228) is returned by every `DidMethod.create` implementation. e.g.

```typescript
const did: PortableDid = await DidDhtMethod.create({ publish: true })
```

when writing code where you want to use the actual did's URI, it ends up looking something like this:

```typescript
const vc = await VerifiableCredential.create({
  type    : 'TBDeveloperCredential',
  subject : did.did,
  issuer  : did.did,
  data    : {
    username: 'moegrammer'
  }
})
```

`did.did` can be a bit confusing. i think changing `did.uri` would make things a bit more clear with respect to when someone is referring the string representation of a did

### LINKED ISSUE #364: Update the `@web5/dids` package to use `@web5/crypto@0.3.0`
Work to be done:
- [ ] Update `@web5/dids` to use `@web5/crypto@0.3.0` now that https://github.com/TBD54566975/web5-js/issues/271 has been completed.

### LINKED ISSUE #368: Update `did:dht` implementation based on spec changes
Several DID DHT spec changes were merged recently: https://github.com/TBD54566975/did-dht-method/pull/86

The JS `did:dht` implementation needs to be updated to reflect the changes.

> [!IMPORTANT]  
> This should be merged _AFTER_ the [DID DHT server implementation](https://github.com/TBD54566975/did-dht-method/tree/main/impl) and TBD hosted node is updated.

frankhinek:
### Summary

The primary goal of this PR is to refactor the `@web5/dids` package to use the refactored `@web5/crypto` package that substantively changed how key management was approached.  In addition to making key management integral to the DID method implementations, it also changes the DID method API to include methods for instantiating DID objects from existing DIDs/keys.  The same pattern is being implemented across `web5-js`, `web5-kt`, `web5-go`, etc.

* closes #263
* closes #329
* closes #364
* closes #274
* closes #368


packages/crypto/src/primitives/xchacha20-poly1305.ts
```diff
@@ -132,27 +132,25 @@ export class XChaCha20Poly1305 {
    * ```
    *
    * @param params - The parameters for the decryption operation.
-   * @param params.data - The encrypted data to decrypt, represented as a Uint8Array.
+   * @param params.data - The encrypted data to decrypt including the authentication tag,
+   *                      represented as a Uint8Array.
    * @param params.key - The key to use for decryption, represented in JWK format.
    * @param params.nonce - The nonce used during the encryption process.
-   * @param params.tag - The authentication tag generated during encryption.
```
nitro-neal, line 138: tag example needs to be removed from example comment code

packages/crypto/src/primitives/xchacha20-poly1305.ts
```diff
@@ -195,17 +193,15 @@ export class XChaCha20Poly1305 {
     data: Uint8Array;
     key: Jwk;
     nonce: Uint8Array;
-  }): Promise<{ ciphertext: Uint8Array, tag: Uint8Array }> {
+  }): Promise<Uint8Array> {
+  // }): Promise<{ ciphertext: Uint8Array, tag: Uint8Array }> {
```
nitro-neal, line 197:
```suggestion

```

packages/crypto/src/primitives/xchacha20-poly1305.ts
frankhinek: Good catch.  Thanks!

Addressed in [38ba056](https://github.com/TBD54566975/web5-js/pull/393/commits/38ba056ca3bfc9486f8e1045d245f66f98f31fc0)

packages/crypto/src/primitives/xchacha20-poly1305.ts
frankhinek: Good catch.  Thanks!

Addressed in [38ba056](https://github.com/TBD54566975/web5-js/pull/393/commits/38ba056ca3bfc9486f8e1045d245f66f98f31fc0)

packages/dids/src/did-uri.ts
```diff
@@ -0,0 +1,186 @@
...
+   */
+  static parse(didUriString: string): DidUri | null {
+    // Return null if the input string is empty or not provided.
+    if (!didUriString) return null;
+
+    // Execute the regex pattern on the input string to extract URI components.
+    const match = DidUri.DID_URI_PATTERN.exec(didUriString);
+
+    // If the pattern does not match, or if the required groups are not found, return null.
+    if (!match || !match.groups) return null;
```
nitro-neal, line 150: No error message or exception? Just returning null if did does not parse?

packages/dids/src/methods/did-dht.ts
```diff
@@ -0,0 +1,1568 @@
...
+      keyManager,
+      verificationMethods: options.verificationMethods ?? []
+    });
+
+    // Create the DID object from the generated key material, including DID document, metadata,
+    // signer convenience function, and URI.
+    const did = await DidDht.fromPublicKeys({ keyManager, options, ...keySet });
+
+    // By default, publish the DID document to a DHT operations endpoint unless explicitly disabled.
+    did.metadata.published = options.publish ?? true
```
nitro-neal, line 533: what happens when someone is on a plane and we try to publish?
timeout exception?

packages/dids/src/did-uri.ts
frankhinek: Yes, this is intentional.  It matches the behavior of the old `parseDid` function it replaces and is checked appropriately for null everywhere it is used.  There are typically specific errors that are returned in DID Resolution Results and this simplifies the logic to return errors without having to add countless try/catch.

packages/dids/src/methods/did-dht.ts
```diff
@@ -0,0 +1,1568 @@
...
+      ? await this.publish({ did, gatewayUri: options.gatewayUri })
+      : false;
+
+    return did;
+  }
+
+  /**
+   * Instantiates a `Did` object for the DID DHT method from a given {@link PortableDid}.
```
nitro-neal, line 541: PortableDid

packages/dids/src/methods/did-dht.ts
```diff
@@ -0,0 +1,1568 @@
...
+    }
+
+    // Read the Fetch Response stream into a byte array.
+    const messageBytes = await response.arrayBuffer();
+
+    if (messageBytes.byteLength < 72) {
+      throw new DidError(DidErrorCode.InvalidDidDocumentLength, `Pkarr response must be at least 72 bytes but got: ${messageBytes.byteLength}`);
+    }
+
+    if (messageBytes.byteLength > 1072) {
```
nitro-neal, line 979: do you need >= 1072 ?

"Pkarr response exceeds 1000 byte limit"
I guess its 72 bytes for base did and then 1000 bytes for other ?

packages/dids/src/methods/did-dht.ts
```diff
@@ -0,0 +1,1568 @@
...
+      dnsAnswerRecords.push({
+        type : 'TXT',
+        name : '_typ._did.',
+        ttl  : DNS_RECORD_TTL,
+        data : `id=${typeIntegers.join(VALUE_SEPARATOR)}`
+      });
+    }
+
+    // Add a DNS TXT record for the root record.
+    dnsAnswerRecords.push({
```
nitro-neal, line 1323: Does this need to be in an else statment? Otherwise you can write 2 txt records.

Is that expected?

packages/dids/src/methods/did-dht.ts
```diff
@@ -0,0 +1,1568 @@
...
+    // Verify that the DID URI is valid.
+    if (!(parsedDid && parsedDid.method === DidDht.methodName)) {
+      throw new DidError(DidErrorCode.InvalidDid, `Invalid DID URI: ${didUri}`);
+    }
+
+    // Decode the method-specific identifier from z-base-32 to a byte array.
+    let identityKeyBytes: Uint8Array | undefined;
+    try {
+      identityKeyBytes = Convert.base32Z(parsedDid.id).toUint8Array();
+    } catch { /* Capture error */ }
```
nitro-neal, line 1474: Do you want to throw new DidError here?

packages/dids/src/methods/did-dht.ts
frankhinek: Yes.  The Mainline DHT max payload is 1000 bytes but `messageBytes` includes a 64 byte signature + 8 byte sequence number

packages/dids/src/methods/did-dht.ts
frankhinek: As currently implemented, if the gateway/relay cannot be reached when attempting to publish an exception will be raised.

One option is to pass `publish: false` so that publishing isn't attempting (when calling `create()` or `fromKeys()`)... but that assumes the application is able to detect a lack of network connectivity and adapt.

One of the downsides is that if a DID is being created for the first time (calling `create()` to generate keys), if an exception is thrown during publishing, those keys will have already been created and stored in the `KeyManager` store.

I went back and forth on this when initially implementing, but another approach is to catch any errors that occur during publishing and return `did.metadata.published: false`.  This way, the caller can detect and attempt to re-publish if they wish... and at least they have the DID URI and key references.

What are your thoughts?

packages/dids/src/methods/did-dht.ts
frankhinek: The error is thrown a few lines down.  This was partly there as a type guard in a slightly different prior implementation.

I thought of a way to refactor and reduce some duplication so pushing up a commit.

Thanks for the second pair of eyes! 😄

packages/dids/src/methods/did-dht.ts
frankhinek: I might be missing the issue or question.  There are potentially many TXT records for a given `did:dht` DID (alsoKnownAs, controller, one for each verification method, one for each service, etc.).

The line you highlighted is adding the final root record, of which there is only one.  However, it is a required record so every DHT DID will have at least 1 TXT record for the root record.

packages/dids/src/methods/did-web.ts
```diff
@@ -0,0 +1,96 @@
...
+    if (parsedDid.method !== DidWeb.methodName) {
+      return {
+        ...EMPTY_DID_RESOLUTION_RESULT,
+        didResolutionMetadata: { error: 'methodNotSupported' }
+      };
+    }
+
+    // Replace ":" with "/" in the identifier and prepend "https://" to obtain the fully qualified
+    // domain name and optional path.
+    let baseUrl = `https://${parsedDid.id.replace(/:/g, '/')}`;
```
ianpatton, line 63: Would be nice to use http:// if the uri is localhost, for local dev purposes

packages/dids/tests/methods/did-web.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+
+      const resolutionResult = await DidWeb.resolve('did:web:non-existent-domain.com');
+
+      expect(resolutionResult.didResolutionMetadata.error).to.equal('notFound');
+
+      fetchStub.restore();
+    });
+  });
+
+  describe('Web5TestVectorsDidJwk', () => {
```
nitro-neal, line 40:
```suggestion
  describe('Web5TestVectorsDidWeb', () => {
```

packages/dids/tests/fixtures/test-vectors/did-resolver/resolve.json
nitro-neal, line 1: empty file, should remove

packages/dids/src/resolver/resolver-cache-level.ts
```diff
@@ -0,0 +1,163 @@
...
+   * @returns The cached DID resolution result or undefined if not found or expired.
+   */
+  async get(did: string): Promise<DidResolutionResult | void> {
+    try {
+      const str = await this.cache.get(did);
+      const cacheWrapper: CacheWrapper = JSON.parse(str);
+
+      if (Date.now() >= cacheWrapper.ttlMillis) {
+        // defer deletion to be called in the next tick of the js event loop
+        this.cache.nextTick(() => this.cache.del(did));
```
nitro-neal, line 105: very cool

packages/dids/src/methods/did-dht.ts
```diff
@@ -0,0 +1,1550 @@
...
+   * Converts a DNS packet to a DID document according to the DID DHT specification.
+   *
+   * @see {@link https://did-dht.com/#dids-as-dns-records | DID DHT Specification, § DIDs as DNS Records}
+   *
+   * @param params - The parameters to use when converting a DNS packet to a DID document.
+   * @param params.didUri - The DID URI of the DID document.
+   * @param params.dnsPacket - The DNS packet to convert to a DID document.
+   * @returns A promise that resolves to a DID document.
+   */
+  private static async fromDnsPacket({ didUri, dnsPacket }: {
```
nitro-neal, line 1046: this is pretty awesome! There is not lib for this? 

If not we can eventually extract this out and make our own lib

nitro-neal (approved):
Wow great job @frankhinek ! 🥇 
I look forward to using this, I think it makes a lot of sense. Lots of new code and hard work.

mistermoe (approved):

packages/dids/tests/fixtures/test-vectors/did-resolver/resolve.json
frankhinek: Great catch @nitro-neal !  I was originally going to add a separate test vector but realized I could use the existing `DidJwk` vectors to serve the same purpose.

I'll remove and push up the commit.

packages/dids/tests/methods/did-web.spec.ts
frankhinek: Needle in a haystack catch.  Nice!

packages/dids/src/methods/did-dht.ts
frankhinek: There's a DNS packet lib that we use for encoding to DNS wire format.  It is used by our DID DHT implementation.  The `fromDnsPacket()` and `toDnsPacket()` implementations in `did-dht.ts` are specific to the encoding/decoding of DID DHT documents.

packages/dids/src/methods/did-dht.ts
frankhinek: Its a great point and the more I think about it... the more I think we ought to look at a results object for DID registration (`create`, `update`, and `deactivate` operations).

I'll leave it out for now and start a design doc to propose an approach.

Early thought is something that looks like: https://github.com/decentralized-identity/did-registration/blob/main/spec/spec.md#didregistrationmetadata

PR #401, opened 2024-02-05
Refactor Credential Signing

nitro-neal:
This change slightly refactors credential signing and verifying. 

Instead of `PortableDid` we now use `BearerDid`

```javascript
import { DidKey } from '@web5/dids';
const issuer: BearerDid = await DidKey.create();

const vcJwt = await vc.sign({ did: issuer });
```

This change also:
* Changes jwt signing to use bearer did's `signer` function 
* Updates all unit tests to use `BearerDid`
* Updates to the following versions
    "@web5/common": "0.2.3",
    "@web5/crypto": "0.4.0",
    "@web5/dids": "0.4.0"


packages/credentials/package.json
```diff
@@ -75,9 +75,9 @@
   },
   "dependencies": {
     "@sphereon/pex": "2.1.0",
-    "@web5/common": "0.2.2",
-    "@web5/crypto": "0.2.4",
-    "@web5/dids": "0.2.4"
+    "@web5/common": "0.2.3",
```
nitro-neal, line 78: These are the latest versions right?

nitro-neal:
Should I bump the version of the credentials package to "version": "0.4.2", ?

packages/credentials/package.json
frankhinek: Yes -- `0.4.0` for both dids and crypto -- `0.2.3` for common.

packages/credentials/src/jwt.ts
```diff
@@ -189,18 +146,11 @@ export class Jwt {
...
-    const algorithmId = `${decodedJwt.header.alg}:${publicKeyJwk['crv'] || ''}`;
-    if (!(algorithmId in Jwt.algorithms)) {
-      throw new Error(`Verification failed: ${algorithmId} not supported`);
-    }
-
-    const { signer, options: signatureAlgorithm } = Jwt.algorithms[algorithmId];
-
-    const isSignatureValid = await signer.verify({
-      algorithm : signatureAlgorithm!,
+    const crypto = new CryptoApi();
```
nitro-neal, line 149: Should we just have is initialized once at the top?

packages/credentials/src/jwt.ts
```diff
@@ -136,14 +99,8 @@ export class Jwt {
     const toSign = `${base64UrlEncodedHeader}.${base64UrlEncodedPayload}`;
     const toSignBytes = Convert.string(toSign).toUint8Array();
 
-    const algorithmId = `${header.alg}:${privateKeyJwk['crv'] || ''}`;
-    if (!(algorithmId in Jwt.algorithms)) {
-      throw new Error(`Signing failed: ${algorithmId} not supported`);
-    }
-
-    const { signer, options: signatureAlgorithm } = Jwt.algorithms[algorithmId];
+    const signatureBytes = await signer.sign({data: toSignBytes});
```
frankhinek, line 102:
```suggestion
    const signatureBytes = await signer.sign({ data: toSignBytes });
```

packages/credentials/src/jwt.ts
frankhinek: sounds good

frankhinek (approved):
Looks good other than a few changes noted above and updating the test

mistermoe (approved):
cc: @angiejones

frankhinek:
> Should I bump the version of the credentials package to "version": "0.4.2", ?

Yes, we might as well publish a release today so that `tbdex` repos can use it.

PR #413, opened 2024-02-15
Fix handling of custom service properties

frankhinek:
This PR will:
- Enable the use of custom service properties such as `enc` and `sig` in the examples below:

    ```ts
    {
      id              : 'did:dht:5cahcfh3zh8bqd5cn3y6inoea1b3d6kh85rjksne9e5dcyrc1ery#dwn',
      type            : 'DecentralizedWebNode',
      serviceEndpoint : 'https://example.com/dwn',
      enc             : '#enc',
      sig             : '#sig',
    }
    ```

    ```ts
    {
      id              : 'did:dht:5cahcfh3zh8bqd5cn3y6inoea1b3d6kh85rjksne9e5dcyrc1ery#dwn',
      type            : 'DecentralizedWebNode',
      serviceEndpoint : 'https://example.com/dwn',
      sig             : ['#auth', '#assert'],
    }
    ```

- Add test coverage for handling string and array custom service properties.

### DIFF
--- a/packages/dids/src/methods/did-dht.ts
+++ b/packages/dids/src/methods/did-dht.ts
@@ -966,7 +966,7 @@ export class DidDhtDocument {
    * @returns A Promise resolving to a {@link DidResolutionResult} object containing the DID
    *          document and its metadata.
    */
-  private static async fromDnsPacket({ didUri, dnsPacket }: {
+  public static async fromDnsPacket({ didUri, dnsPacket }: {
     didUri: string;
     dnsPacket: Packet;
   }): Promise<DidResolutionResult> {
@@ -1055,11 +1055,16 @@ export class DidDhtDocument {
           // The service endpoint can either be a string or an array of strings.
           const serviceEndpoint = se.includes(VALUE_SEPARATOR) ? se.split(VALUE_SEPARATOR) : se;
 
+          // Convert custom property values to either a string or an array of strings.
+          const serviceProperties = Object.fromEntries(Object.entries(customProperties).map(
+            ([k, v]) => [k, v.includes(VALUE_SEPARATOR) ? v.split(VALUE_SEPARATOR) : v]
+          ));
+
           // Initialize the `service` array if it does not already exist.
           didDocument.service ??= [];
 
           didDocument.service.push({
-            ...customProperties,
+            ...serviceProperties,
             id   : `${didUri}#${id}`,
             type : t,
             serviceEndpoint
@@ -1115,7 +1120,7 @@ export class DidDhtDocument {
    * @param params.didMetadata - The DID metadata to include in the DNS packet.
    * @returns A promise that resolves to a DNS packet.
    */
-  private static async toDnsPacket({ didDocument, didMetadata }: {
+  public static async toDnsPacket({ didDocument, didMetadata }: {
     didDocument: DidDocument;
     didMetadata: DidMetadata;
   }): Promise<Packet> {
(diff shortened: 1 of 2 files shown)


nitro-neal (approved):
looks good!

csuwildcat (approved):

decentralgabe:
> This PR will:
> 
> * Enable the use of custom service properties such as `enc` and `sig` in the examples below:
>   ```ts
>   {
>     id              : 'did:dht:5cahcfh3zh8bqd5cn3y6inoea1b3d6kh85rjksne9e5dcyrc1ery#dwn',
>     type            : 'DecentralizedWebNode',
>     serviceEndpoint : 'https://example.com/dwn',
>     enc             : '#enc',
>     sig             : '#sig',
>   }
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   ```ts
>   {
>     id              : 'did:dht:5cahcfh3zh8bqd5cn3y6inoea1b3d6kh85rjksne9e5dcyrc1ery#dwn',
>     type            : 'DecentralizedWebNode',
>     serviceEndpoint : 'https://example.com/dwn',
>     sig             : ['#auth', '#assert'],
>   }
>   ```
> * Add test coverage for handling string and array custom service properties.

I believe it should look like

```ts
   {
     id              : 'did:dht:5cahcfh3zh8bqd5cn3y6inoea1b3d6kh85rjksne9e5dcyrc1ery#dwn',
     type            : 'DecentralizedWebNode',
     serviceEndpoint : 'https://example.com/dwn',
     enc             : 'enc',
     sig             : 'sig',
   }
```

```ts
   {
     id              : 'did:dht:5cahcfh3zh8bqd5cn3y6inoea1b3d6kh85rjksne9e5dcyrc1ery#dwn',
     type            : 'DecentralizedWebNode',
     serviceEndpoint : 'https://example.com/dwn',
     sig             : ['auth', 'assert'],
   }
```

since the identifiers are `enc` or `sig` not `#enc` or `#sig`

packages/dids/src/methods/did-dht.ts
```diff
@@ -1188,13 +1193,14 @@ export class DidDhtDocument {
     didDocument.service?.forEach((service, index) => {
       const dnsRecordId = `s${index}`;
       serviceIds.push(dnsRecordId);
-      const serviceId = service.id.split('#').pop()!; // Remove fragment prefix, if any.
```
decentralgabe, line 1191: so this should be returned

frankhinek: @decentralgabe -- nothing changed in that line -- the `extractDidFragment()` does the exact same thing as `.split('#').pop()`

frankhinek:
> I believe it should look like
> 
> ```ts
>    {
>      id              : 'did:dht:5cahcfh3zh8bqd5cn3y6inoea1b3d6kh85rjksne9e5dcyrc1ery#dwn',
>      type            : 'DecentralizedWebNode',
>      serviceEndpoint : 'https://example.com/dwn',
>      enc             : 'enc',
>      sig             : 'sig',
>    }
> ```
> 
> ```ts
>    {
>      id              : 'did:dht:5cahcfh3zh8bqd5cn3y6inoea1b3d6kh85rjksne9e5dcyrc1ery#dwn',
>      type            : 'DecentralizedWebNode',
>      serviceEndpoint : 'https://example.com/dwn',
>      sig             : ['auth', 'assert'],
>    }
> ```
> 
> since the identifiers are `enc` or `sig` not `#enc` or `#sig`

With custom properties it doesn't seem like we should have special handling for the values.  At least that's how I've been thinking about it.

In other words, there's nothing to say that the values of custom properties are referring to an `id` of a verification method.  They could refer to almost anything.

decentralgabe:
@frankhinek agree if they do not reference values in the doc itself then no issue

frankhinek:
> @frankhinek agree if they do not reference values in the doc itself then no issue

Are you thinking there should be logic to check the DID document to try and figure out whether the values are referring to other document properties?

decentralgabe:
the guidance is currently only present around keys, but I'd like to update this to recommend fully qualified identifiers for all properties https://did-dht.com/#representing-keys, I'll open an issue for this.

generally I think we should avoid fragments with # since - at least to me - they only make sense in the context of a full identifier

PR #414, opened 2024-02-15
Fix several minor issues in `@web5/dids`

frankhinek:
This PR will:
- Correct typos in comments.
- Improve type inference in `DidMethodApi` generics.
- Add `getSigningMethod()` to `DidMethodApi` to ensure all implementations include it.
- Make the `verificationMethods` type in `DidJwk` and `DidKey` to be consistent with `DidDht`.
- Bump `@web5/dids` to `0.4.1` so that the changes can be published to NPM Registry.

> [!IMPORTANT]
> Should be merged _after_ #413

### DIFF
--- a/packages/dids/src/methods/did-method.ts
+++ b/packages/dids/src/methods/did-method.ts
@@ -82,9 +82,9 @@ export interface DidCreateVerificationMethod<TKms> extends Pick<Partial<DidVerif
  * @typeparam O - The type of the options used for creating the DID.
  */
 export interface DidMethodApi<
-    TDid extends BearerDid,
-    TOptions extends DidCreateOptions<TKms>,
-    TKms extends CryptoApi | undefined = undefined
+    TKms extends CryptoApi | undefined = CryptoApi,
+    TDid extends BearerDid = BearerDid,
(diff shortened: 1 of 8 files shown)


mistermoe (approved):

packages/dids/src/methods/did-method.ts
```diff
@@ -82,9 +82,9 @@ export interface DidCreateVerificationMethod<TKms> extends Pick<Partial<DidVerif
  * @typeparam O - The type of the options used for creating the DID.
  */
 export interface DidMethodApi<
-    TDid extends BearerDid,
-    TOptions extends DidCreateOptions<TKms>,
-    TKms extends CryptoApi | undefined = undefined
+    TKms extends CryptoApi | undefined = CryptoApi,
```
nitro-neal, line 85: any breaking changes in downstream packages expected?

nitro-neal (approved):

packages/dids/src/methods/did-method.ts
frankhinek: Nope -- this reduces the chances of any breaking changes by using a default that can be used (as opposed to `undefined`)

PR #415, opened 2024-02-15
Add TTL Cache to `@web5/common` package

frankhinek:
This PR will:
- Add a 0 dependency, small footprint (`25.2 KB` unpacked), well used (> 500K weekly downloads), TypeScript TTL Cache to the `@web5/common` package.
- Bumps `@web5/common` to `0.2.4` so that the package can be published and consumed by other Web5 packages.

### Context

This cache can be used as an in-memory TTL cache option for other Web5 packages such as `@web5/dids` (e.g., `DidResolver`), `@web5/agent`, etc.  Adding to the `common` package to ensure consistent use across packages.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1358,6 +1358,14 @@
         "url": "https://github.com/chalk/strip-ansi?sponsor=1"
       }
     },
+    "node_modules/@isaacs/ttlcache": {
+      "version": "1.4.1",
+      "resolved": "https://registry.npmjs.org/@isaacs/ttlcache/-/ttlcache-1.4.1.tgz",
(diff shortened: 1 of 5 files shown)


nitro-neal (approved):

mistermoe (commented):
This is cool. Should we include an example somewhere of how to use it with our resolver?

frankhinek:
> This is cool. Should we include an example somewhere of how to use it with our resolver?

Absolutely.  Was planning on adding the example usage in `@web5/dids`.

But also happy to add somewhere else if you had a thought here.

mistermoe (approved):

PR #430, opened 2024-02-23
Add crypto utility function and use in BearerDid

### LINKED ISSUE #418: Correct `BearerDid` documentation
While working on an unrelated activity I noticed that the description for `BearerDid` had not been updated to reflect changes that enable its `export()` method to work with key managers, like AWS KMS, that don't support exporting private keys.

The comments for `BearerDid` should be corrected: https://github.com/TBD54566975/web5-js/blob/1d595e116234fe7a70f758774e943049fe065662/packages/dids/src/bearer-did.ts#L81-L110

frankhinek:
This PR will:
- Move a crypto utility function from within `BearerDid` to the `@web5/crypto` package.
- Update `BearerDid` to use the utility function.
- Update the docs for `BearerDid` to close #418.

> [!NOTE]
> The package versions for `@web5/crypto` and `@web5/dids` are not incremented in this PR as this change is rather minimal, has no immediate impact on external API consumers, and consequently, isn't worth publishing a new release.
> A future PR will bump the versions and publish new releases as the utility function added to `@web5/crypto` is consumed in other packages in this monorepo (which is already the case for a branch that is still a WIP.

### DIFF
--- a/packages/crypto/src/utils.ts
+++ b/packages/crypto/src/utils.ts
@@ -1,3 +1,5 @@
+import type { Jwk } from './jose/jwk.js';
+
 import { crypto } from '@noble/hashes/crypto';
 import { randomBytes as nobleRandomBytes } from '@noble/hashes/utils';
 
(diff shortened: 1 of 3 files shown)


packages/crypto/src/utils.ts
```diff
@@ -63,6 +65,54 @@ export function checkValidProperty(params: {
...
+ * const algorithm = getJoseSignatureAlgorithmFromPublicKey(publicKey);
+ * console.log(algorithm); // Output: "EdDSA"
+ * ```
+ *
+ * @param publicKey - A JWK containing the `alg` and/or `crv` properties.
+ * @returns The name of the algorithm associated with the key.
+ * @throws Error if the algorithm cannot be determined from the provided input.
+ */
+export function getJoseSignatureAlgorithmFromPublicKey(publicKey: Jwk): string {
+  const curveToJoseAlgorithm: Record<string, string> = {
```
jiyoontbd, line 94: typescript q: is `Record` better than `Map` in this context?

packages/crypto/src/utils.ts
```diff
@@ -1,3 +1,5 @@
+import type { Jwk } from './jose/jwk.js';
```
jiyoontbd, line 1: is there some initiative to eject from jose/jwk in web5-js, or is ejecting from dependency effort just for web5-kt?

jiyoontbd (approved):

packages/crypto/src/utils.ts
frankhinek: Good question.  There are performance advantages and you do get nicer lookups but slightly more verbose initialization and reverse lookups.  

But practically speaking, this data structure is so small (and will remain so given how few JOSE DSAs exist) that either approach is acceptable.  I opted to take the slightly more terse route in this case.

packages/crypto/src/utils.ts
frankhinek: Just web5-let in the context of JOSE (JWK, JWS, JWE, JWT) and ejecting from external deps.

There is an initiative that underway to consolidate the aforementioned JOSE functionality into the `@web5/crypto` package.  That entails migrating JWT from `@web5/credentials` and completing implementation of JWS and JWE.  JWK already lives in `crypto`.  I’ve got a draft of Compact and Flattened JWE just about done so will open a PR for those soon.

packages/crypto/src/utils.ts
```diff
@@ -63,6 +65,54 @@ export function checkValidProperty(params: {
...
+  if (publicKey.alg && Object.values(curveToJoseAlgorithm).includes(publicKey.alg)) {
+    return publicKey.alg;
+  }
+
+  // If the key contains a `crv` property, return the corresponding algorithm.
+  if (publicKey.crv && Object.keys(curveToJoseAlgorithm).includes(publicKey.crv)) {
+    return curveToJoseAlgorithm[publicKey.crv];
+  }
+
+  throw new Error(`Unable to determine algorithm based on provided input: alg=${publicKey.alg}, crv=${publicKey.crv}`);
```
KendallWeihe, line 113: Maybe we should include the curves we support here and/or in the docs?

KendallWeihe (approved):
seems right to me 👍

packages/crypto/src/utils.ts
diehuxx: @jiyoontbd Additional typescript neckbeardery if you're curious / didn't already know.

`Record` is pure typescript in the sense that javascript does not have a notion of `Record`. A `Record` is just an plain old javascript object, but with some detail provided for the typescript compiler.

`Map` exists in javascript at runtime, so it's not "just" a typescript thing. `Map` is a [class that you can instantiate directly](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) using `new Map()` and use instance methods `.get()` or `.set()`.

Like Frank mentioned, the difference in performance at runtime between a plain old javascript object and a `Map` is negligible for our purposes. [This stackoverflow answer](https://stackoverflow.com/questions/18541940/map-vs-object-in-javascript) provides some color on when `Map` is more useful.

KendallWeihe:
Only nitpicky feedback item, so take it or leave it, would've been to do this one https://github.com/TBD54566975/web5-js/issues/418 is a separate PR. Reasoning being, it's a disparate matter to the other two items in this PR so caused a slight disruption when reviewing, but also it may just be because of my naivete to this code base.

packages/crypto/src/utils.ts
```diff
@@ -63,6 +65,54 @@ export function checkValidProperty(params: {
...
+    'Ed25519'   : 'EdDSA',
+    'P-256'     : 'ES256',
+    'P-384'     : 'ES384',
+    'P-521'     : 'ES512',
+    'secp256k1' : 'ES256K',
+  };
+
+  // If the key contains an `alg` property that matches a JOSE registered algorithm identifier,
+  // return its value.
+  if (publicKey.alg && Object.values(curveToJoseAlgorithm).includes(publicKey.alg)) {
```
diehuxx, line 104: Nit clarification: Previously, we didn't check that the alg is in `curveToJoseAlgorithm` values. This seems correct to me, but also took me a second to notice while reviewing. Did we previously want to return "unsupported" algorithms or was that just an oversight that we're not correcting?

diehuxx (approved):
This change seems pretty innocuous since we're just moving the location of existing functionality. I added a question, but feel free to merge as-is.

packages/crypto/src/utils.ts
frankhinek: I may be missing the question.  Unsupported algorithms won't be returned and an error will be thrown.

packages/crypto/src/utils.ts
frankhinek: Thanks @KendallWeihe.  Added the list of supported algorithm identifiers and curve names to the error.

PR #433, opened 2024-02-29
`@web5/agent` DWN + Web5 RPC Clients

LiranCohen:
In this PR:

- organize rpc clients into `agent/prototyping/clients` until they have a permanent home
- copy `JsonRpcSocket` class from `@web5/dwn-server`, using `isomorphic-ws` to allow for isomorphic socket client.
- create `WebSocketDwnRpcClient` and `WebSocketWeb5RpcClient` to support web5 requests over sockets.


shamilovtim (approved):
Looks good. Avoiding reviewing methods that were just moved from one place to another since that's not in scope of the PR.

Only real comment is that the transportProtocols method seems possibly superfluous (are we worried someone would import HTTPWeb5RpcClient and try to use it to do something that's not HTTP?)

frankhinek (approved):

PR #436, opened 2024-03-04
Add hermit and update GH actions

phoebe-lew:
closes #376

PR #437, opened 2024-03-06
fix a couple of actions and bump

shamilovtim:
- I noticed the `node:test` action failed on main from #436 where there was a bump to Node v20. 

- Also noticed that a couple of the CI steps were still using node v18, so I have updated those to also use the same setup as what was introduced in #436. 

- I also bumped the actions versions to newer ones

- There is a perceptible regression in node performance between node v18 and node v20 which caused the  `SyncManagerLevel` test timeout to be exceeded. By bumping the timeout of the `agent` test suite I was able to account for the larger test time. 

- I was able to parallelize `test-with-node` which reduced that CI step's runtime by about 8%

- The mocha junit reporter is obscuring test results on CI so failures are unknown at the moment. I think this is an outstanding problem, and the junit reporter is doing more harm than good.

### DIFF
--- a/.github/workflows/tests-ci.yml
+++ b/.github/workflows/tests-ci.yml
@@ -55,8 +55,11 @@ jobs:
           node node_modules/@web5/dwn-server/dist/esm/src/main.js &
           echo "DWN_SERVER_BACKGROUND_PROCESS=$!" >> $GITHUB_ENV
 
+      - name: Build tests for all packages
+        run: pnpm --recursive --stream --sequential build:tests:node
+
       - name: Run tests for all packages
-        run: npm run test:node --ws -- --color
+        run: pnpm --recursive --stream exec c8 mocha -- --color --reporter mocha-junit-reporter --reporter-options mochaFile=./results.xml
         env:
           TEST_DWN_URL: http://localhost:3000
 
--- a/.github/workflows/tests-ci.yml
+++ b/.github/workflows/tests-ci.yml
@@ -59,7 +59,7 @@ jobs:
         run: pnpm --recursive --stream --sequential build:tests:node
 
       - name: Run tests for all packages
-        run: npm run test:node --ws -- --color
+        run: pnpm --recursive --stream exec c8 mocha -- --color --reporter mocha-junit-reporter --reporter-options mochaFile=./results.xml
         env:
           TEST_DWN_URL: http://localhost:3000
 
--- a/.github/workflows/tests-ci.yml
+++ b/.github/workflows/tests-ci.yml
@@ -56,7 +56,7 @@ jobs:
           echo "DWN_SERVER_BACKGROUND_PROCESS=$!" >> $GITHUB_ENV
 
       - name: Run tests for all packages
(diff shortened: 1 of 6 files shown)


.github/workflows/tests-ci.yml
```diff
@@ -65,7 +56,7 @@ jobs:
           echo "DWN_SERVER_BACKGROUND_PROCESS=$!" >> $GITHUB_ENV
 
       - name: Run tests for all packages
-        run: pnpm --recursive --no-stream --sequential test:node -- --color --reporter mocha-junit-reporter --reporter-options mochaFile=./results.xml
+        run: npm run test:node --ws -- --color
```
shamilovtim, line 59: debugging this. something broke here with the upgrade to node (or hermit).

packages/agent/.mocharc.json
```diff
@@ -1,5 +1,6 @@
 {
   "enable-source-maps": true,
   "exit": true,
-  "spec": ["tests/compiled/**/*.spec.js"]
+  "spec": ["tests/compiled/**/*.spec.js"],
+  "timeout": 5000
```
frankhinek, line 5:
```suggestion
  "spec": ["tests/compiled/**/*.spec.js"]
```
```suggestion
  "spec": ["tests/compiled/**/*.spec.js"],
  "timeout": 5000
```

.github/workflows/tests-ci.yml
```diff
@@ -64,8 +55,11 @@ jobs:
           node node_modules/@web5/dwn-server/dist/esm/src/main.js &
           echo "DWN_SERVER_BACKGROUND_PROCESS=$!" >> $GITHUB_ENV
 
+      - name: Build tests for all packages
+        run: pnpm --recursive --stream --sequential build:tests:node
+
       - name: Run tests for all packages
-        run: pnpm --recursive --no-stream --sequential test:node -- --color --reporter mocha-junit-reporter --reporter-options mochaFile=./results.xml
+        run: pnpm --recursive --stream exec c8 mocha -- --color --reporter mocha-junit-reporter --reporter-options mochaFile=./results.xml
```
frankhinek, line 62:
```suggestion
        run: npm run test:node --ws -- --color
```

.github/workflows/tests-ci.yml
```diff
@@ -64,8 +55,11 @@ jobs:
           node node_modules/@web5/dwn-server/dist/esm/src/main.js &
           echo "DWN_SERVER_BACKGROUND_PROCESS=$!" >> $GITHUB_ENV
 
+      - name: Build tests for all packages
+        run: pnpm --recursive --stream --sequential build:tests:node
+
       - name: Run tests for all packages
-        run: pnpm --recursive --no-stream --sequential test:node -- --color --reporter mocha-junit-reporter --reporter-options mochaFile=./results.xml
+        run: npm run test:node --ws -- --color
```
frankhinek, line 62: Revert to Tim's changes now that timeouts are resolved
```suggestion
        run: pnpm --recursive --stream exec c8 mocha -- --color --reporter mocha-junit-reporter --reporter-options mochaFile=./results.xml
```

frankhinek (approved):

csuwildcat (approved):

PR #416, opened 2024-02-15
Setup Changesets

leordev:
- Introduces and setup Changesets
- Whole information and details were added on the `CONTRIBUTING.md` file
- Automation for preview packages on every merged changeset to main under the tag `next`
- Since `@web5/api` requires a release orchestration with the community and a series of checks through a release template checklist, version bumps on changesets PRs are disabled by default; when ready to release trigger the release workflow manually with the API flag or just label the Version Packages PR with the `api-release` label. (Also added to the contributing file)

TODO before merging:
- [x] Create the `api-release` label

TODO after merging:
- [ ] Install Changesets bot

Working evidence in my forked repo:
- Git Tags and Releases with Changelog notes:
  - https://github.com/leordev/web5-js/tags
  - https://github.com/leordev/web5-js/releases
- npm.js versions with preview next tags being published: 
  - https://www.npmjs.com/package/@leordev-web5/api?activeTab=versions
- Changesets Version Package PR example
  -  https://github.com/leordev/web5-js/pull/3
- PR with Changesets bot notification:
  - https://github.com/leordev/web5-js/pull/2

PR #438, opened 2024-03-07
publish snapshot releases on every merge to main

finn-tbd:
This is part of a project being undertaken by OSP to help new code make it's way to TBD's downstream projects faster. It creates a snapshot release for each merge to main, which is published to our Artifactory* registry. Once this is done, downstream projects (ie. tbdex-js and developer.tbd.website) can pull in new web5-js features before a real release has been produced by depending on the snapshot version.

It creates snapshot release for any packages that have changed since the last snapshot version. It does this by counting the number of commits in each package since the last time the `package.json` version was changed, and producing a snapshot release name in the format `<version>-SNAPSHOT.<commit count>` (for example, there have been 55 commits that touched `packages/api` since it was bumped to `0.8.4`, so it will produce snapshot version `0.8.4-SNAPSHOT.55`). EDIT: this format has changed slightly to include the sha of the last commit as well, see thread. It checks if a snapshot with this version has already been published, and if not it builds and publishes the snapshot release.

Once that has been published, someone writing documentation for developer.tbd.website can set the example to depend on `@web5/api` version `0.8.4-SNAPSHOT.55` and write sample code that can be validated against that snapshot.

*Artifactory: behaves like npmjs.com, but isn't real npmjs.com so the high number of releases this creates will not spam up our npmjs.com page.

### DIFF
--- a/.github/workflows/release-snapshot.yml
+++ b/.github/workflows/release-snapshot.yml
@@ -1,0 +1,100 @@
+name: release snapshot to jfrog
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+# Allow only one concurrent deployment, but do NOT cancel in-progress runs as
+# we want to allow these release deployments to complete.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+  id-token: write
+
+jobs:
+  publish-npm:
+    name: Snapshot Publish
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        package:
+          [
+            "agent",
+            "api",
+            "common",
+            "credentials",
+            "crypto",
+            "crypto-aws-kms",
+            "dids",
+            "identity-agent",
+            "proxy-agent",
+            "user-agent",
+          ]
+
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 #v4.1.1
+        with:
+          fetch-depth: 0
+
+      # https://cashapp.github.io/hermit/usage/ci/
+      - name: Init Hermit
+        uses: cashapp/activate-hermit@v1
+        with:
+          cache: "true"
+
+      - uses: jfrog/setup-jfrog-cli@v4
+        with:
+          version: latest
+          oidc-provider-name: github # must match the OpenID Connect name from https://blockxyz.jfrog.io/ui/admin/configuration/integrations
+        env:
+          JF_URL: https://blockxyz.jfrog.io
+
+      - name: Publish @web5/${{ matrix.package }} snapshot
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.npm_token}}
+        run: |
+          set -exuo pipefail
+
+          package_name="${{ matrix.package }}"
+          cd "packages/${package_name}"
+
+          base_version=$(jq -r .version package.json)
+
+          # I'm not seeing a great way to determine the commit of the last release of a given package, so I'm using a not-so-great way
+          version_line=$(grep -n "\"version\": \"${base_version}\"" package.json | cut -d: -f1) # determine which line in package.json specifies the version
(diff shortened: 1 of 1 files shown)


.github/workflows/release-snapshot.yml
```diff
@@ -0,0 +1,100 @@
...
+            "identity-agent",
+            "proxy-agent",
+            "user-agent",
+          ]
+
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 #v4.1.1
+        with:
+          fetch-depth: 0
```
leordev, line 44: isnt it going to fetch the entire tree? if so, do we need it?

leordev: ah nevermind... you need it to determine if the version was bumped 😢

leordev (approved):
lgtm! I think I might be able to simplify things after merging #416  (the `next` publication could publish to jfrog too! and changesets is smart enough to do the diff logic)

frankhinek:
@finn-tbd Is it possible to publish the short SHA in the version `<version>-SNAPSHOT.<short-SHA>` rather than commit count?

If no, how does a consumer of the published release know which commit it is associated with?  I'm imagining a scenario where a downstream consumer reports they ran into an issue/bug with `0.8.4-SNAPSHOT.55`.  During troubleshooting, how would we determine which commit in the `main` branch this release was cut from?  Is the only way to manually look up the commit for `0.8.4` and then count 55 commits later?

finn-tbd:
It's possible, but that would break npm's versioning scheme. in particular, it would make it unable to determine which is the latest snapshot. If that's something we're super concerned about, we could combine the two approaches (ie. `<version>-SNAPSHOT.<count>-<sha>`) and probably have comparable versions, but i'm 100% sure how npm's version comparing works. Should I implement the combined approach?

frankhinek:
> It's possible, but that would break npm's versioning scheme. in particular, it would make it unable to determine which is the latest snapshot.

Can you elaborate on the concern?

Based on `npm publish` command in this PR it appears the desired outcome is that if someone runs `npm install @web5/<package-name> --registry https://blockxyz.jfrog.io/artifactory/api/npm/tbd-oss-snapshots-npm/` without specifying a specific version that `latest` will be the most recent commit to main?

Do we really want `latest` to be the latest commit to main or should `latest` be the most recent version that went through the normal release process?

finn-tbd:
> Based on npm publish command in this PR it appears the desired outcome is that if someone runs npm install @web5/<package-name> --registry https://blockxyz.jfrog.io/artifactory/api/npm/tbd-oss-snapshots-npm/ without specifying a specific version that latest will be the most recent commit to main?

correct

> Do we really want latest to be the latest commit to main or should latest be the most recent version that went through the normal release process?

I would expect the latest version in the snapshot repository to be the latest snapshot. If we just used the SHA and had no count, it would still install a snapshot release, it just wouldn't be the latest one. I'm not 100% sure how it would pick but i think it would be the (alphabetically) highest SHA, so closest to `fffffff`.

i've updated it to include the SHA in the snapshot version, so version numbers are computed like this:

```bash
snapshot_version="${base_version}-SNAPSHOT.${commits_since_version_bump}-${last_commit_to_package:0:7}"
```

for example, `@web5/crypto`'s latest snapshot number is `0.4.0-SNAPSHOT.97-7442465`

.github/workflows/release-snapshot.yml
```diff
@@ -0,0 +1,101 @@
...
+
+      # https://cashapp.github.io/hermit/usage/ci/
+      - name: Init Hermit
+        uses: cashapp/activate-hermit@v1
+        with:
+          cache: "true"
+
+      - uses: jfrog/setup-jfrog-cli@v4
```
frankhinek, line 52: Since the security tools added by OSP raise security issues as a result of unpinned actions, please use the hash and include a comment with which version it is.

frankhinek (changes requested):

.github/workflows/release-snapshot.yml
finn-tbd: whoops, thanks for catching that. Also pinned `cashapp/activate-hermit` above

frankhinek: no worries.  thanks!  nice work @finn-tbd

frankhinek (approved):

PR #439, opened 2024-03-08
Update alpha-npm.yml to fix publish

nitro-neal:
successful runs with npm:
image

unsuccessful run with pnpm:
image

there is no option `--no-git-tag-version` in the documentation https://pnpm.io/cli/publish#--tag-tag

I removed it, but please check if this is how we want alpha releases handled, or if we ant to create a custom tag (of if there is an option for removing the tag that I'm not aware of)

### DIFF
--- a/.github/workflows/alpha-npm.yml
+++ b/.github/workflows/alpha-npm.yml
@@ -62,7 +62,7 @@ jobs:
         run: |
           for package in $PACKAGES; do
             cd packages/$package
-            pnpm publish --tag alpha --no-git-tag-version --access public --provenance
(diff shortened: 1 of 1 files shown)


shamilovtim (approved):
Curious the intent behind --no-git-tag-version. Maybe it makes sense to use npm only for this command?

frankhinek (approved):

frankhinek:
> Curious the intent behind --no-git-tag-version. Maybe it makes sense to use npm only for this command?

IIRC it was to ensure that git tags weren't created in the repo for alphas... but that doesn't seem to be an issue with `pnpm`.

Thanks @nitro-neal !

PR #441, opened 2024-03-11
Fix path Esbuild config to run all browser tests

frankhinek:
## Summary

This PR will:
- Change `entryPoints : ['./tests/*.spec.*'],` in `build/esbuild-tests.cjs` to `entryPoints : ['./tests/**/*.spec.*'],` to run tests in sub-directories during Web Test Runner browser tests.
- Remove `chai-as-promised` usage without imports in `@web5/dids` test spec files to resolve failing browser test runs.
- Skip running AES 192-bit tests in Chrome browser (_see context below_).
- Ensure that the `tagLength` property is never passed as `undefined` to the Web Crypto API to resolve a test that only fails in the Chrome browser.

## Context

Prior to this PR, any tests that were in a sub-directory of `tests/` were _not_ being built (transpiled to JavaScript), and as a consequence, the tests weren't run by Web Test Runner.

Once all of the browser tests were back to running, a few failures surfaced in the `dids` and `crypto` packages that required changes to get the tests back to passing:

- The most notable of these was changes due to Chrome browser taking a different approach to implementing the Web Crypto API than WebKit / Gecko: Chrome/Blink supports encryption/decryption with AES 128- and 256-bit keys but not 192-bit.  Since 192-bit length is part of the NIST standard, rather than removing support from `@web5/crypto` entirely, the test is skipped in Chrome. This ensures developers building backend services (i.e., non-browser runtime) can decrypt ciphertext that was produced by another lib. In practice, its not a major issue since AES 256-bit encryption/decryption is preferred by many modern apps and is hardware accelerated by CPU AES-NI in modern browsers.
- The `dids` package change was minimal in that two test spec files contained references to `chai-as-promised` extensions but were lacking the import.  Rather than add the imports, the usage of `chai-as-promised` was removed.  This is part of an ongoing effort to remove usage of this extension package to make migrating off Mocha/Chai easier at some point in the future.


packages/dids/package.json
```diff
@@ -101,7 +100,6 @@
     "@web/test-runner-playwright": "0.11.0",
     "c8": "9.0.0",
     "chai": "4.3.10",
-    "chai-as-promised": "7.1.1",
```
shamilovtim, line 104: I encourage moving off of this package in the long term because it breaks compat with codemods and the larger js echosystem like Jest, Vite or Bun's test runner

shamilovtim (approved):

packages/dids/package.json
frankhinek: 💯 Have been slowing trying to remove it to make a migration to a different test framework easier.

.github/workflows/tests-ci.yml
```diff
@@ -59,7 +59,8 @@ jobs:
         run: pnpm --recursive --stream --sequential build:tests:node
 
       - name: Run tests for all packages
-        run: pnpm --recursive --stream exec c8 mocha -- --color --reporter mocha-junit-reporter --reporter-options mochaFile=./results.xml
+        # run: pnpm --recursive --stream exec c8 mocha -- --color --reporter mocha-junit-reporter --reporter-options mochaFile=./results.xml
```
shamilovtim, line 62: I think in the future if you just delete `--reporter mocha-junit-reporter --reporter-options mochaFile=./results.xml` you can still debug without switching to the npm command but I aim to have the fix PR up today for a reference for @finn-tbd

csuwildcat (approved):

LiranCohen (approved):

PR #446, opened 2024-03-12
Fix VSCode workspace paths after switch to pnpm

frankhinek:
## Summary

This PR will:
- Switch from `root` to package-specific workspace paths for the Mocha binary which restores ability to use VSCode debugging.

## Context

When the Web5 JS monorepo was migrated from `npm` to `pnpm` one of the consequences is that root `node_modules` directory no longer contains the kitchen sink of every dependency used by every package in the monorepo.  As a result, the path reference in each package's `launch.json` for the Mocha binary is no longer valid.

### DIFF
--- a/packages/agent/.vscode/launch.json
+++ b/packages/agent/.vscode/launch.json
@@ -8,7 +8,7 @@
       "type": "node",
       "request": "launch",
       "name": "test:node",
-      "runtimeExecutable": "${workspaceFolder:root}/node_modules/.bin/mocha",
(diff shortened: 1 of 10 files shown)


csuwildcat (approved):

thehenrytsai (approved):
🐐

LiranCohen (approved):

shamilovtim (approved):

mistermoe (approved):

PR #448, opened 2024-03-14
Status List Credential Implementation

nitro-neal:
`StatusListCredential` represents a digitally verifiable status list credential according to the [W3C Verifiable Credentials Status List v2021](https://www.w3.org/community/reports/credentials/CG-FINAL-vc-status-list-2021-20230102/).

When a status list is published, the result is a verifiable credential that encapsulates the status list.

Implements status list credential `create` and `validateCredentialInStatusList`

`create`
```
  /**
   * Create a [StatusListCredential] with a specific purpose, e.g., for revocation.
   *
   * @param statusListCredentialId The id used for the resolvable path to the status list credential [String].
   * @param issuer The issuer URI of the credential, as a [String].
   * @param statusPurpose The status purpose of the status list cred, eg: revocation, as a [StatusPurpose].
   * @param issuedCredentials The credentials to be included in the status list credential, eg: revoked credentials, list of type [VerifiableCredential].
   * @returns A special [VerifiableCredential] instance that is a StatusListCredential.
   * @throws Error If the status list credential cannot be created.
   *
   * Example:
   * ```
      StatusListCredential.create({
        statusListCredentialId : 'https://statuslistcred.com/123',
        issuer                 : issuerDid.uri,
        statusPurpose          : StatusPurpose.REVOCATION,
        issuedCredentials      : [credWithCredStatus]
      })
   * ```
   */
   ```

`validateCredentialInStatusList`
```
  /**
   * Validates if a given credential is part of the status list represented by a [VerifiableCredential].
   *
   * @param credentialToValidate The [VerifiableCredential] to be validated against the status list.
   * @param statusListCredential The [VerifiableCredential] representing the status list.
   * @returns A [Boolean] indicating whether the `credentialToValidate` is part of the status list.
   *
   * This function checks if the given `credentialToValidate`'s status list index is present in the expanded status list derived from the `statusListCredential`.
   *
   * Example:
   * ```
   * const isRevoked = StatusListCredential.validateCredentialInStatusList(credentialToCheck, statusListCred);
   * ```
   */
```

Closes: https://github.com/TBD54566975/web5-js/issues/380

### CODE REVIEW

File: packages/credentials/tests/status-list-credential.spec.ts [hunk 1]
```diff
@@ -0,0 +1,232 @@
...
+        type             : 'StreetCred',
+        issuer           : issuerDid.uri,
+        subject          : subjectDid.uri,
+        data             : new StreetCredibility('high', true),
+        credentialStatus : credentialStatus
+      });
+
+      const credWithStatusContexts = credWithCredStatus.vcDataModel['@context'] as string[];
+
+      expect(credWithStatusContexts.includes('https://www.w3.org/2018/credentials/v1')).to.be.true;
```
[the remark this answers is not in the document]
nitro-neal: fixed

packages/credentials/tests/status-list-credential.spec.ts
```diff
@@ -0,0 +1,232 @@
...
+      expect(credWithStatusContexts.includes('https://w3id.org/vc/status-list/2021/v1')).to.be.true;
```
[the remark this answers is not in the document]
nitro-neal: fixed

packages/credentials/src/status-list-credential.ts
```diff
@@ -0,0 +1,235 @@
...
+export enum StatusPurpose {
+  REVOCATION = 'revocation',
+  SUSPENSION = 'suspension',
+}
+
+/**
+ * The size of the bitstring in bits.
+ * The bitstring is 16KB in size.
+ */
+const BITSTRING_SIZE = 16 * 1024 * 8; // 16KB in bits
```
decentralgabe, line 18: may want to consider this as an overridable option in creating the list

decentralgabe (approved):
nice job on this. the most important thing is to get this into web5 specs with some tests to make sure we can check statuses consistently across impls

nitro-neal:
Adding test vectors here https://github.com/TBD54566975/web5-spec/pull/151/files

packages/credentials/src/status-list-credential.ts
```diff
@@ -0,0 +1,238 @@
...
+import { VerifiableCredential, DEFAULT_VC_CONTEXT, DEFAULT_VC_TYPE, VcDataModel } from './verifiable-credential.js';
+import type { ICredentialStatus} from '@sphereon/ssi-types';
+import { Convert } from '@web5/common';
+
+export const DEFAULT_STATUS_LIST_VC_CONTEXT = 'https://w3id.org/vc/status-list/2021/v1';
+export const DEFAULT_STATUS_LIST_VC_TYPE = 'StatusList2021Credential';
+
+export enum StatusPurpose {
+  REVOCATION = 'revocation',
+  SUSPENSION = 'suspension',
```
thehenrytsai, line 12: Unconventional to use all caps for enum values, consider:
```suggestion
  Revocation = 'Revocation',
  Suspension = 'Suspension',
```

or if these values are predefined to be lower case per spec:

```suggestion
  revocation = 'revocation',
  suspension = 'suspension',
```

thehenrytsai: Handled in https://github.com/TBD54566975/web5-js/pull/486.

packages/credentials/src/status-list-credential.ts
```diff
@@ -0,0 +1,238 @@
...
+   * ```
+      StatusListCredential.create({
+        statusListCredentialId : 'https://statuslistcred.com/123',
+        issuer                 : issuerDid.uri,
+        statusPurpose          : StatusPurpose.REVOCATION,
+        issuedCredentials      : [credWithCredStatus]
+      })
+   * ```
+   */
+  public static create(options: StatusListCredentialCreateOptions): VerifiableCredential {
```
thehenrytsai, line 75: Design: from status list lifecycle management and dev experience standpoint, it would be more useful if:

1. There is an option that takes in an existing status list to be operated on.
    ie. Revoke a VC from an existing status list. This would be the main line scenario of status list usage.
2. It would be nice to support toggle from 1 to 0 also.
    ie. Reactivate a suspension of VC.

packages/credentials/src/status-list-credential.ts
```diff
@@ -0,0 +1,238 @@
...
+    const decompressedData = pako.inflate(compressedData);
+
+    // Convert the decompressed data into an array of "0" or "1" strings
+    const bitstringArray: number[] = [];
+    decompressedData.forEach(byte => {
+      for (let i = 7; i >= 0; i--) {
+        const bit = (byte >> i) & 1;
+        bitstringArray.push(bit);
+      }
+    });
```
thehenrytsai, line 234: This is pretty inefficient: will iterate ~160,000 times minimum (assuming 16KB size) when can be done with an O(1) look up.

Consider an alternative imple by changing the method from `bitstringExpansion()` to something like `getBit(bitArray, bitIndex): boolean`, something like:

``` ts
      // Find the byte index, and bit index within the byte.
    const byteIndex = Math.floor(bitIndex / 8);
    const bitIndexWithinByte = bitIndex % 8;

    const byte = decompressedData[byteIndex];

    const bitInteger = (byte >> bitIndexWithinByte) & 1; // Remove all preceding bits using bit shift, then remove trailing bits using &.
    return (bitInteger === 1);
```

packages/credentials/src/status-list-credential.ts
```diff
@@ -0,0 +1,238 @@
...
+   * @returns A [Boolean] indicating whether the `credentialToValidate` is part of the status list.
+   *
+   * This function checks if the given `credentialToValidate`'s status list index is present in the expanded status list derived from the `statusListCredential`.
+   *
+   * Example:
+   * ```
+   * const isRevoked = StatusListCredential.validateCredentialInStatusList(credentialToCheck, statusListCred);
+   * ```
+   */
+  public static validateCredentialInStatusList(
```
thehenrytsai, line 113: Design: there is probably a combination of misunderstanding of the spec, bad naming of the method, or bad ChatGPT documentation generation 😆 :

When the value is is 1 in the given index position, it does not mean the credential is "in the status list" as the method name and documentation states. It just means the credential is revoked. So we ought to rename the method to something like `isCredentialRevoked()`, though I generally prefer having method checking for the happy case, in which case `isCredentialValid()` or `validateCredentialStatus()`.

packages/credentials/tests/status-list-credential.spec.ts
```diff
@@ -0,0 +1,268 @@
...
+      });
+
+      expect(statusListCredential).not.be.undefined;
+      expect(statusListCredential.subject).to.equal('revocation-id');
+
+      const credentialSubject = statusListCredential.vcDataModel.credentialSubject as any;
+      expect(credentialSubject['type']).to.equal('StatusList2021');
+      expect(credentialSubject['statusPurpose']).to.equal('revocation');
+
+      // TODO: Check encoding across other sdks and spec - https://github.com/TBD54566975/web5-kt/issues/97
```
thehenrytsai, line 141: Remove or reference an active issue.

thehenrytsai (changes requested):
Hey @nitro-neal, I have a PR for your PR #486 which has some less-controversial refactoring and renaming, we should get that merged into your PR before you taking on further changes based on my comments to avoid conflicts galore!

packages/credentials/src/status-list-credential.ts
nitro-neal: genius! yes much better

packages/credentials/src/status-list-credential.ts
nitro-neal: Yup! You are right that is probably true the api needs to be updated

The current api copies that of the kt and the old go impl, The status list -is- regenerated every time there is an update, so its basically creating a new one I guess was the thought here.

I agree the api needs to be updated to be easier to grok. I'll create an iussue

packages/credentials/src/status-list-credential.ts
nitro-neal: I agree it needs to be updated along with the other api naming changes, This is the same naming as web5-kt and yea will want to do a light design and update the api naming across all sdks

A problem here is that the function name has to be generic because it can be revoked, suspended, annulled, withdrawn ...

But yes will add to the issue

nitro-neal:
Updated and merged in your updates.

thehenrytsai (approved):
🚀

packages/credentials/src/status-list-credential.ts
thehenrytsai, line 1: Fix the tbdocs warning yo!

thehenrytsai (commented):
Looks good, except for that typedoc warning... I thought I turned it into an error instead of warning. Maybe this is still running off of old code... will investigate if it continues to be like this.

nitro-neal:
spec update
https://github.com/TBD54566975/web5-spec/pull/151

thehenrytsai (approved):
🚀

Issue #447, opened 2024-03-14
`PresentationExchange.selectCredentials` returns all vcJwts when none match


kirahsapong:
Given an array of JWTs where NONE of the credentials satisfy the given PD, ALL credentials are returned from `PresentationExchange.selectCredentials()`. This only happens when NONE of the credentials match. If there's at least one credential that satisfies the PD, then `selectCredentials()` works.

**Steps to reproduce**
1. Create a `presentationDefinition` with input descriptors
2. Create an array of `vcJwts` where none of the `vcJwts` satisfies the `presentationDefinition`
3. Call `PresentationExchange.selectCredentials(vcJwts, presentationDefinition)`

**Current behaviour**
Return value will be the original array of vcJwts

**Expected behaviour**
Return value should be an array with a size of 0

_**Sidenote (if you have time)**_: `satisfiesPresentationDefinition` should return a boolean instead of throwing when vcJwts do not satisfy. Throwing causes code to stop executing which doesn't work for iterating over a list to filter out presentation definitions

nitro-neal:
pr here - https://github.com/TBD54566975/web5-js/pull/450

PR #450, opened 2024-03-18
Fix select credential bug

nitro-neal:
fixes - https://github.com/TBD54566975/web5-js/issues/447

### DIFF
--- a/packages/credentials/src/presentation-exchange.ts
+++ b/packages/credentials/src/presentation-exchange.ts
@@ -34,7 +34,7 @@ export class PresentationExchange {
     const selectResults: SelectResults = this.pex.selectFrom(presentationDefinition, vcJwts);
 
     // If errors exist in the results object the credentials provided didn't satisfy the requirements in the Presentation Definition
-    if(selectResults.errors?.length) {
+    if(selectResults.errors?.length !== 0) {
       return [];
     }
 
--- a/packages/credentials/src/presentation-exchange.ts
+++ b/packages/credentials/src/presentation-exchange.ts
@@ -32,7 +32,13 @@ export class PresentationExchange {
   }): string[] {
     this.resetPex();
     const selectResults: SelectResults = this.pex.selectFrom(presentationDefinition, vcJwts);
-    return Array.from(new Set(selectResults.verifiableCredential as string[] ?? []));
+
+    // If errros the credentials provided didn't satisfy the requirement defined presentationDefinition object
+    if(selectResults.errors?.length) {
+      return [];
+    } else {
+      return Array.from(new Set(selectResults.verifiableCredential as string[] ?? []));
+    }
   }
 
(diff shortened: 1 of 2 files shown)


packages/credentials/tests/presentation-exchange.spec.ts
```diff
@@ -58,6 +65,24 @@ describe('PresentationExchange', () => {
       expect(actualSelectedVcJwts).to.deep.equal([btcCredentialJwt]);
     });
 
+    it('should return 0 verifiable credentials when they dont match the pd', async () => {
```
frankhinek, line 68: minor nit but probably doesn't hurt to spell out:
```suggestion
    it('should return 0 verifiable credentials when they dont match the presentation definition', async () => {
```

packages/credentials/src/presentation-exchange.ts
```diff
@@ -32,7 +32,13 @@ export class PresentationExchange {
   }): string[] {
     this.resetPex();
     const selectResults: SelectResults = this.pex.selectFrom(presentationDefinition, vcJwts);
-    return Array.from(new Set(selectResults.verifiableCredential as string[] ?? []));
+
+    // If errros the credentials provided didn't satisfy the requirement defined presentationDefinition object
```
frankhinek, line 36: Small typo but possible to provide a better explanation based on this detail: https://github.com/TBD54566975/web5-js/issues/447
```suggestion
    // If errros the credentials provided didn't satisfy the requirement defined presentationDefinition object
```

frankhinek (approved):

packages/credentials/src/presentation-exchange.ts
```diff
@@ -32,6 +32,12 @@ export class PresentationExchange {
   }): string[] {
     this.resetPex();
     const selectResults: SelectResults = this.pex.selectFrom(presentationDefinition, vcJwts);
+
+    // If errors exist in the results object the credentials provided didn't satisfy the requirements in the Presentation Definition
+    if(selectResults.errors?.length) {
```
diehuxx, line 37: super nit: despite using JS for years i still get confused about falsy values. do you mind doing `!== 0`?

packages/credentials/tests/presentation-exchange.spec.ts
```diff
@@ -50,6 +50,13 @@ describe('PresentationExchange', () => {
       });
     });
 
+    it('should evaluate credentials without any errors or warnings', async () => {
+      PresentationExchange.satisfiesPresentationDefinition({
```
diehuxx, line 54: nit: make the test description a little more descriptive.
is this test just testing that it doesn't throw as long as the cred satisfies the PD or is it meant to show that it never throws even when the cred fails to satisfy the PD?

diehuxx (approved):

PR #452, opened 2024-03-19
Fix scorecard

finn-tbd:
This seemingly fixes the scorecard analysis job. I'm not clear exactly what broke but it seems like the version we pinned can no longer talk to their online service. There are two changes:
* bump the version of the action to latest
* run the action in PRs to prove it works

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -37,7 +37,7 @@ jobs:
           persist-credentials: false
 
       - name: "Run analysis"
-        uses: ossf/scorecard-action@e38b1902ae4f44df626f11ba0734b14fb91f8f86 # v2.1.2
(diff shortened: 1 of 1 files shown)


.github/workflows/scorecard.yml
```diff
@@ -13,6 +13,8 @@ on:
     - cron: '27 22 * * 4'
   push:
     branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
```
leordev, line 17: Problem to introduce this in PRs is that we are going to introduce the red failure check that might be happening for every PR, even if its not introduced in the corresponding PR (thus pushing that bad habit of making devs to ignore the CI checks that we wanted to fix initially in our pipelines).

finn-tbd: removed pr trigger

leordev (approved):

frankhinek (approved):

PR #454, opened 2024-03-20
Add DID resolution types and rename UniversalResolver

frankhinek:
## Summary

This PR will:
- Add `DidResolver` and `DidUrlDereferencer` types to `@web5/dids` package.
- Rename the concrete implementation that supports resolving and dereferencing multiple DID methods to `UniversalResolver`.

## Context

Packages that depend on `DidResolver` throw errors when module resolution attempts to import different versions of the `@web5/dids` package.  This typically occurs whenever the package version is bumped in one but not both packages.  This happens, for example, whenever the [alpha publishing](https://github.com/TBD54566975/web5-js/actions/workflows/alpha-npm.yml) workflow is run.

Providing a `DidResolver` type that the downstream packages can depend on resolves this and has already been tested.

Additional details [here](https://github.com/TBD54566975/web5-spec/issues/137).

### DIFF
--- a/packages/dids/src/resolver/universal-resolver.ts
+++ b/packages/dids/src/resolver/universal-resolver.ts
@@ -1,27 +1,22 @@
-import type { KeyValueStore } from '@web5/common';
-
 import type { DidMethodResolver } from '../methods/did-method.js';
+import type { DidResolver, DidResolverCache, DidUrlDereferencer } from '../types/did-resolution.js';
 import type { DidDereferencingOptions, DidDereferencingResult, DidResolutionOptions, DidResolutionResult, DidResource } from '../types/did-core.js';
 
 import { Did } from '../did.js';
 import { DidErrorCode } from '../did-error.js';
 import { DidResolverCacheNoop } from './resolver-cache-noop.js';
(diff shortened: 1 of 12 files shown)


packages/dids/src/resolver/universal-resolver.ts
```diff
@@ -153,7 +136,6 @@ export class DidResolver {
   /**
    * Dereferences a DID (Decentralized Identifier) URL to a corresponding DID resource.
    *
```
thehenrytsai, line 138: Unless the typedoc tooling will complain, consider removing duplicated documentation that is already provided in the interface, VSC automatically infers from the interface.

thehenrytsai (approved):
🚀 🚀 🚀

packages/dids/src/resolver/universal-resolver.ts
frankhinek: @thehenrytsai This is for the purposes of generating the API reference documentation rather than VSCode.

And to avoid the docs reporting from reporting these:
Screenshot 2024-03-20 at 4 54 18 PM

mistermoe (approved):

PR #467, opened 2024-03-27
Add OSP Reusable Workflow for Scans

leordev:
Similar to https://github.com/TBD54566975/developer.tbd.website/pull/1348

It also include the OSSF Scan checks.

PS: Note that it's fine to reuse a workflow without pinning the dependency because OSSF checks for the nested workflow. (In this case we are setting to main branch `TBD54566975/open-source-programs/.github/workflows/security.yml@main` ).


thehenrytsai (approved):
@leordev Nice work for abstracting these workflows away from the dependent repos!

PR #468, opened 2024-03-27
Remove typedoc warnings in credentials package

nitro-neal:
This adds comments to remove the warnings that were generated by 

`pnpx typedoc --validation --entryPoints src/index.ts`

in the credentials package

### DIFF
--- a/packages/credentials/src/verifiable-credential.ts
+++ b/packages/credentials/src/verifiable-credential.ts
@@ -7,7 +7,9 @@ import { Jwt } from './jwt.js';
 import { SsiValidator } from './validators.js';
 import { getCurrentXmlSchema112Timestamp, getXmlSchema112Timestamp } from './utils.js';
 
+/** The default Verifiable Credential context. */
 export const DEFAULT_VC_CONTEXT = 'https://www.w3.org/2018/credentials/v1';
+/** The default Verifiable Credential type. */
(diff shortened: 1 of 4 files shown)


jiyoontbd (approved):
wao, docs!

jiyoontbd:
@nitro-neal is it possible that the long wall tbdoc warnings that show up on every single PR can be fixed by running the cmd in the other packages?

nitro-neal:
Yup the goal is to clean up all docs, I think agent and api is in the works and I can tackle crypto-aws-kms

diehuxx (approved):
Screenshot 2024-03-28 at 11 17 08 AM
docs

packages/credentials/src/verifiable-credential.ts
```diff
@@ -278,8 +292,11 @@ export class VerifiableCredential {
     validatePayload(vcTyped);
 
     return {
+      /** The issuer of the VC. */
```
shamilovtim, line 295: These are already defined up above. Are the docs really complaining about every single key of a returned object? If so that probably needs to get fixed on the docs check

shamilovtim: Oh these are separate and unrelated to the class, nevermind.

shamilovtim (approved):

PR #486, opened 2024-04-05
Updates to status list 2021 PR

thehenrytsai:


packages/credentials/src/status-list-credential.ts
```diff
@@ -1,47 +1,57 @@
...
  * @param statusPurpose The status purpose of the status list cred, eg: revocation, as a [StatusPurpose].
- * @param issuedCredentials The credentials to be included in the status list credential, eg: revoked credentials, list of type [VerifiableCredential].
+ * @param revokedCredentials The credentials to be included in the status list credential, eg: revoked credentials, list of type [VerifiableCredential].
  */
 export type StatusListCredentialCreateOptions = {
   statusListCredentialId: string,
   issuer: string,
   statusPurpose: StatusPurpose,
-  issuedCredentials: VerifiableCredential[]
+  revokedCredentials: VerifiableCredential[]
```
nitro-neal, line 36: A credential can be revoked, suspeneded, or any other noun actually. Status Purpose should actually probably just be a string

packages/credentials/src/status-list-credential.ts
```diff
@@ -67,15 +77,15 @@ export class StatusListCredential {
...
+        revokedCredentials      : [credWithCredStatus]
       })
    * ```
    */
   public static create(options: StatusListCredentialCreateOptions): VerifiableCredential {
-    const { statusListCredentialId, issuer, statusPurpose, issuedCredentials } = options;
-    const statusListIndexes: string[] = this.prepareCredentialsForStatusList(statusPurpose, issuedCredentials);
-    const bitString = this.bitstringGeneration(statusListIndexes);
+    const { statusListCredentialId, issuer, statusPurpose, revokedCredentials } = options;
+    const indexOfCredentialsToRevoke: number[] = this.validateStatusListEntryIndexesAreAllUnique(statusPurpose, revokedCredentials);
```
nitro-neal, line 87:
```suggestion
    const indexsOfCredentialsToRevoke: number[] = this.validateStatusListEntryIndexesAreAllUnique(statusPurpose, revokedCredentials);
```

packages/credentials/src/status-list-credential.ts
```diff
@@ -1,47 +1,57 @@
...
 /**
- * The StatusList2021Entry instance representing the core data model of a vc status list 2021.
+ * Credential status lookup information included in a Verifiable Credential that supports status lookup.
+ * Data model dictated by the Status List 2021 spec.
  *
- * @see {@link https://www.w3.org/community/reports/credentials/CG-FINAL-vc-status-list-2021-20230102/ | Status List 2021 Entry}
+ * @see {@link https://www.w3.org/community/reports/credentials/CG-FINAL-vc-status-list-2021-20230102/#example-example-statuslist2021credential | Status List 2021 Entry}
  */
-export interface StatusList2021Entry extends ICredentialStatus {
+export interface CredentialStatusModel {
```
nitro-neal, line 45: Following our spec,  and for clarity I think it should e called `StatusList2021Entry`

https://github.com/TBD54566975/web5-spec/blob/main/spec/vc.md

There is a different credStatusModel that will be for vc data model 2.0

packages/credentials/src/status-list-credential.ts
thehenrytsai: Good point, though the action of suspend or revoke result in the same bit flip to 1, this would probably go away if you decide to take my other PR suggestion of have `revoke()`, for now I'll rename to `credentialToDisable`. Happy to discuss but I don't think `includedCredentials` is it.

packages/credentials/src/status-list-credential.ts
thehenrytsai: Good catch, done!

nitro-neal (approved):

nitro-neal:
Looks good, little nits, and we can sync on some nuances next week if needed. Thanks for this!

packages/credentials/src/status-list-credential.ts
thehenrytsai: Reverted! I had strong opinion on this, but you win on this one! We can chat about it next week.

PR #489, opened 2024-04-11
`@web5/agent` Adding `DwnServerInfo` to RPC Clients

LiranCohen:
- Added a `DwnServerInfo` HTTP client to get info from the `dwn-server`'s `/info` endpoint

```
export type ServerInfo = {
  /** the maximum file size the user can request to store */
  maxFileSize: number,
  /**
   * an array of strings representing the server's registration requirements.
   *
   * ie. ['proof-of-work-sha256-v0', 'terms-of-service']
   * */
  registrationRequirements: string[],
  /** whether web socket support is enabled on this server */
  webSocketSupport: boolean,
}
```

This is helpful for retrieving registration requirements and whether the server supports sockets.

It uses a `TTLCache` memory cache as the currently implemented and default, however additional caches can be easily added if necessary.


packages/agent/src/prototyping/clients/dwn-server-info-cache-level.ts
```diff
@@ -0,0 +1,159 @@
...
+ * @remarks
+ * The LevelDB cache keeps data in memory for fast access and also writes to the filesystem in
+ * Node.js or indexedDB in browsers. Time-to-live (TTL) for cache entries is configurable.
+ *
+ * @example
+ * ```
+ * const cache = new DwnServerInfoCacheLevel({ ttl: '15m' });
+ * ```
+ */
+export class DwnServerInfoCacheLevel implements DwnServerInfoCache {
```
shamilovtim, line 70: Is there a strong reason for having persistence here? Just a very minor concern that we're adding too many persisted caches in this package where it might not be called for. Especially when browser / mobile storage is limited?

packages/agent/src/prototyping/clients/dwn-server-info-cache-level.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  constructor({
+    db,
+    location = 'DATA/DWN_SERVERINFOCACHE',
+    ttl = '15m'
+  }: DwnServerCacheLevelParams = {}) {
+    this.cache = db ?? new Level<string, string>(location);
+    this.ttl = ms(ttl);
+  }
+
```
shamilovtim, line 85: the API looks very familiar. do we already have something like this elsewhere in the project?

shamilovtim (commented):
Could the two caches somehow reuse the `MemoryStore` or `LevelStore` interfaces? It seems like we've written that same API a handful of time already in web5 SDK

LiranCohen:
@shamilovtim good call outs on the unnecessary caching implementations. We can add ones if there is a need but I think a memory TTL cache is all that we currently need.

> Could the two caches somehow reuse the MemoryStore or LevelStore interfaces? It seems like we've written that same API a handful of time already in web5 SDK

The `MemoryStore` is simply a map where as the `TTLCache` has a bit more logic to evict data once a size is hit.
I would like to create an issue to create an `AsyncTTLCache` or something of that nature that would fit an async interface better, effectively a generic version of what `DwnServerInfoCacheMemory` is doing. We could create that in `@web5/common` and then replace some of these other wrappers that are just doing the same thing by extending the `AsyncTTLCache`.

Let me know what you think about the approach and I'll create an issue.

shamilovtim (approved):

packages/agent/tests/prototyping/clients/dwn-server-info-cache.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+      const resultAfterDelete = await cache.get(key1);
+      expect(resultAfterDelete).to.equal(undefined);
+      const result2AfterDelete = await cache.get(key2);
+      expect(result2AfterDelete).to.equal(undefined);
+    });
+
+    it('returns undefined after ttl', async function () {
+      // skip this test in the browser, sinon fake timers don't seem to work here
+      // with a an await setTimeout in the test, it passes.
+      if (!isNode) {
```
thehenrytsai, line 101: Sanity: this would be the first precedence that it doesn't work in browser right? Have you tried passing `number` instead of `string` to `tickAsync()`?

packages/agent/src/prototyping/clients/dwn-server-info-cache-memory.ts
```diff
@@ -0,0 +1,79 @@
...
+  /**
+   * Optional. The time-to-live for cache entries, expressed as a string (e.g., '1h', '15m').
+   * Determines how long a cache entry should remain valid before being considered expired.
+   *
+   * Defaults to '15m' if not specified.
+   */
+  ttl?: string;
+}
+
+export class DwnServerInfoCacheMemory implements DwnServerInfoCache {
```
thehenrytsai, line 21: Consider turning this into generic, seems low hanging fruit.

```suggestion
export class InMemoryCache<T> implements GenericCache<T> {
```

thehenrytsai (commented):
Looks good! Couple of comments.

packages/agent/src/prototyping/clients/dwn-server-info-cache-memory.ts
shamilovtim: The thing about this is we already have one called MemoryStore which is why his is more specific. Maybe MemoryStoreAsync? Unsure

packages/agent/tests/prototyping/clients/dwn-server-info-cache.spec.ts
LiranCohen: It's really strange that it doesn't work. I just tried a number instead of string and get the same behavior because I didn't think to try that before. But still fails in the browsers.

packages/agent/src/prototyping/clients/dwn-server-info-cache-memory.ts
LiranCohen: @thehenrytsai Yeah @shamilovtim  called out something similar and I was considering this, but figured we could do a separate PR to add `TTLCacheAsync` or something of that sort to `@web5/common` because we are doing this same wrapper in a couple of places to conform with an async interface.

PR #491, opened 2024-04-12
Credential Schema VcDataModel Impl

nitro-neal:
Implement Credential Schema

A credential schema defines the structure and content of the data, enabling verifiers to assess if the data adheres to the established schema.

https://www.w3.org/TR/vc-data-model/#data-schemas

* Add and export CredentialSchema type
* Add optional field to VcCreateOptions
* Update KYC unit test to have latest kyc cred


KendallWeihe (approved):
seems simple enough, and no semver binary incompatibilities 

do you think we should verify the schema during the signing functionality @nitro-neal ?

jiyoontbd:
looks like security audit is failing @nitro-neal  - something about mysql? can you look into this and either bump or remove the dep

nitro-neal:
there is a pr for the vuln here - https://github.com/TBD54566975/dwn-sql-store/pull/25

And yup there will be a future pr about resolving the schema, and failing on verify if the schema does not match.

I think this should go in on our modular approach work to verify, some verifiers may not care about the schema so we need Verify Options to have it be optional

jiyoontbd:
can you explain how fixing the vuln in a different repo would also address it here? is it somehow linked?

jiyoontbd (approved):

frankhinek (approved):

PR #496, opened 2024-04-19
DID Resolution Caching

LiranCohen:
Currently we are running into a problem where a DID's cache TTL is up and it is evicted from the cache, but a subsequent call may return a `notFound` error and it currently caches it.

The goal for this PR is to:
- Not cache results that contain a resolution error.

### DIFF
--- a/packages/dids/src/resolver/universal-resolver.ts
+++ b/packages/dids/src/resolver/universal-resolver.ts
@@ -126,8 +126,10 @@ export class UniversalResolver implements DidResolver, DidUrlDereferencer {
       return cachedResolutionResult;
     } else {
       const resolutionResult = await resolver.resolve(parsedDid.uri, options);
-
-      await this.cache.set(parsedDid.uri, resolutionResult);
+      if (!resolutionResult.didResolutionMetadata.error) {
+        // Cache the resolution result if it was successful.
+        await this.cache.set(parsedDid.uri, resolutionResult);
+      }
 
       return resolutionResult;
     }
--- a/.changeset/tall-toes-remember.md
+++ b/.changeset/tall-toes-remember.md
@@ -1,0 +1,5 @@
+---
+"@web5/dids": patch
+---
+
+Do not cache results that contain a resolution error.
--- a/packages/dids/tests/resolver/universal-resolver.spec.ts
+++ b/packages/dids/tests/resolver/universal-resolver.spec.ts
@@ -19,6 +19,10 @@ describe('UniversalResolver', () => {
       didResolver = new UniversalResolver({ didResolvers: didMethodApis });
     });
 
+    afterEach(() => {
+      sinon.restore();
+    });
+
     it('returns an invalidDid error if the DID cannot be parsed', async () => {
       const didResolutionResult = await didResolver.resolve('unparseable:did');
       expect(didResolutionResult).to.exist;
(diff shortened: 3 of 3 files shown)


csuwildcat (approved):

michaelneale:
yeah - concept makes sense, never really want to cache a negative.

packages/dids/src/types/did-core.ts
```diff
@@ -339,6 +339,11 @@ export type DidResolutionMetadata = {
    */
   error?: string;
 
+  /**
+   * Signals whether or not the resolved DID document is stale in the cache.
+   */
+  isStaleInCache?: boolean;
```
KendallWeihe, line 345: Could we come up with a way of implementing this state which doesn't involve adding a property to `DidResolutionMetadata` which isn't in the specification data model https://www.w3.org/TR/did-core/#did-resolution-metadata?

LiranCohen: Thanks! Great call out!

I extended the `DidResolutionResult` as `CachedDidResolutionResult` and put that property within there.

Let me know what you think about this approach.

packages/dids/src/resolver/universal-resolver.ts
```diff
@@ -121,16 +121,28 @@ export class UniversalResolver implements DidResolver, DidUrlDereferencer {
     }
 
     const cachedResolutionResult = await this.cache.get(parsedDid.uri);
-
     if (cachedResolutionResult) {
+      if (cachedResolutionResult?.isStaleInCache === true) {
+        // if the cache is stale, we would like to attempt refreshing it after returning the existing stale result
+        isomorphicNextTick(() => this.resolveAndCache(parsedDid.uri, resolver, options));
```
KendallWeihe, line 127: haven't been in js land in a while, do we have a potential async issue here in that `resolveAndCache` is async but `isomorphicNextTick` is not, and we want to block until `resolveAndCache` is completed?

also, should `resolveAndCache` return value be set to `cachedResolutionResult`

so something like (probably not totally correct)

```suggestion
        cachedResolutionResult = await isomorphicNextTick(() => this.resolveAndCache(parsedDid.uri, resolver, options));
```

packages/dids/src/types/did-core.ts
KendallWeihe: yeah this is great 👍

packages/dids/src/resolver/universal-resolver.ts
LiranCohen: In this particular case I didn't want to block while resolving + caching as the `dht` resolution can take up to 10 seconds. Talking with @csuwildcat we figure the chance of a stale `did` document being incorrect is probably very slim and is affected by general caching anyway so it's preferable to return a stale result than wait for a new one (which could also fail resolution intermittently).

So that was the purpose of introducing an `isomorphicNextTick` to allow js to trigger this after the function returns without having to block for it.

JS land is tricky with this kind of stuff but this *seems* to be the right way to do it.

KendallWeihe: excellent

KendallWeihe (approved):

packages/dids/src/resolver/universal-resolver.ts
```diff
@@ -121,16 +121,28 @@ export class UniversalResolver implements DidResolver, DidUrlDereferencer {
...
+      }
+
       return cachedResolutionResult;
```
thehenrytsai, line 130: If I am reading this correctly, we alway return the stale data when `isStaleInCache` is `true`. This seems to be to aggressive, imagine there is a DID that is stale in cache for 1 year (extremely example to demonstrate the point) because it is just not used very often. A new resolution resolution will return this 1 year old DID state even though there may have been multiple updates to this DID.

Consider an alternate implementation where there is a timeout for resolving DID (say 5 seconds) when it is stale, use the new state if it is returned, and if nothing/error comes back, it refreshes the TTL using the stale data.

thehenrytsai (commented):
Some initial thoughts. Will read through rest of the PR.

packages/dids/src/resolver/universal-resolver.ts
LiranCohen: Good idea, i'll implement something a little more sane for that.

packages/dids/src/resolver/resolver-cache-level.ts
```diff
@@ -95,19 +95,17 @@ export class DidResolverCacheLevel implements DidResolverCache {
...
    */
-  async get(did: string): Promise<DidResolutionResult | void> {
+  async get(did: string): Promise<CachedDidResolutionResult | void> {
     try {
       const str = await this.cache.get(did);
       const cacheWrapper: CacheWrapper = JSON.parse(str);
 
       if (Date.now() >= cacheWrapper.ttlMillis) {
-        // defer deletion to be called in the next tick of the js event loop
-        this.cache.nextTick(() => this.cache.del(did));
```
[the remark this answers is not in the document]
LiranCohen: Yup you're correct, but I also think this was the case prior if the same DID was only fetched once. I will take a look a bit further and see if we can make sure to evict based on a set size + LRU.

thehenrytsai (approved):
🚀

PR #503, opened 2024-04-29
speed up and rebalance CI flow

shamilovtim:
speeds up and rebalances the CI flow
entire run is now 18% faster

### DIFF
--- a/.github/workflows/tests-ci.yml
+++ b/.github/workflows/tests-ci.yml
@@ -82,11 +82,13 @@ jobs:
           - group: "A"
             packages: "--filter agent"
           - group: "B"
-            packages: "--filter credentials ----filter dids --filter identity-agent"
+            packages: "--filter dids --filter identity-agent"
           - group: "C"
-            packages: "--filter api --filter common"
+            packages: "--filter api"
           - group: "D"
-            packages: "--filter crypto --filter proxy-agent --filter user-agent"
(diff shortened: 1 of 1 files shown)


LiranCohen (approved):
Nice! Speed up in CI is always a win!

For my knowledge, is there any reason not to group all of the various `agent` packages together? Would that make things slower? Just curious of the groupings have some sort of logic around them or if it's mostly based on the speed of the run.

shamilovtim:
> Nice! Speed up in CI is always a win!
> 
> For my knowledge, is there any reason not to group all of the various `agent` packages together? Would that make things slower? Just curious of the groupings have some sort of logic around them or if it's mostly based on the speed of the run.

no reason to group package together based on logic unless the tests are not isolated and need to be run in order. test balance comes from test timings. in better systems than github actions the CI is able to infer test timing from output and auto rebalance. with github actions you have to balance manually.

LiranCohen:
> test balance comes from test timings. 

Yeah that's what i figured.

> in better systems than github actions the CI is able to infer test timing from output and auto rebalance. with github actions you have to balance manually

Glad you're keeping tabs on this! 18% is a huge gain! 🎸

thehenrytsai (approved):
🎸 
🙏 
Massive thanks! QoL improvements are powerful and always welcome!

PR #504, opened 2024-05-01
Updated DID:DHT implementation to be compliant with vector 1

thehenrytsai:
1. Updated DID:DHT implementation to be compliant with vector 1
1. Added self to default owners in CODEOWNERS
1. Opportunistic QoL update: excluded generated files from search results
1. Opportunistic QoL update:: allow one click debugging for DID project

### DIFF
--- a/packages/dids/src/methods/did-dht.ts
+++ b/packages/dids/src/methods/did-dht.ts
@@ -1255,18 +1255,9 @@ export class DidDhtDocument {
     // Add a DNS TXT record for the root record.
     dnsAnswerRecords.push({
       type : 'TXT',
-      name : '_did.',
+      name : '_did.' + DidDhtDocument.getUniqueDidSuffix(didDocument.id) + '.', // name of a Root Record MUST end in `<ID>.`
       ttl  : DNS_RECORD_TTL,
       data : rootRecord.join(PROPERTY_SEPARATOR)
-    });
-
-    // Per the DID DHT specification (https://did-dht.com/#dids-as-dns-records):
-    // Record name of a Root Record or a Authoritative Gateway Records MUST end in `<ID>.`
-    const [, , identifier] = didDocument.id.split(':');
-    dnsAnswerRecords.forEach(record => {
-      if (record.name.startsWith('_did.')) {
-        record.name += identifier + '.';
-      }
     });
 
     // Create a DNS response packet with the authoritative answer flag set.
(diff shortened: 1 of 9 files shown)


packages/dids/tests/methods/did-dht.spec.ts
```diff
@@ -1153,10 +1160,35 @@ describe('DidDhtDocument', () => {
...
+      for (const vector of resolveTestVectors.vectors as any[]) {
         const didResolutionResult = await DidDht.resolve(vector.input.didUri);
         expect(didResolutionResult.didResolutionMetadata.error).to.equal(vector.output.didResolutionMetadata.error);
       }
     }).timeout(30000); // Set timeout to 30 seconds for this test for did:dht resolution timeout test
   });
 });
+
+// vectors come from https://did-dht.com/#test-vectors
+describe('Official DID:DHT Vector tests', () => {
```
nitro-neal, line 1172: ok this needs to be the other way,
Our test vectors look in a hardcoded spot 

Web5TestVectorsDidDht.resolve()
image

so keep the resolve() one the same with the web5-spec test vectors, and then make a new one called something like:

describe('Local DID:DHT resolve vectors, () => {

nitro-neal (approved):

packages/dids/src/methods/did-dht.ts
```diff
@@ -1260,10 +1260,14 @@ export class DidDhtDocument {
...
-    // Origin of all records.
+    // Per the DID DHT specification (https://did-dht.com/#dids-as-dns-records):
+    // Record name of a Root Record or a Authoritative Gateway Records MUST end in `<ID>.`
     const [, , identifier] = didDocument.id.split(':');
-    dnsAnswerRecords.forEach(record => record.name += identifier);
+    dnsAnswerRecords.forEach(record => {
+      if (record.name.startsWith('_did.')) {
+        record.name += identifier + '.';
+      }
+    });
```
diehuxx, line 1270: Instead of appending identifier after creating the `answer`, what do you think of setting it when we create the root record and auth gateway records? For the root record, that would mean updating this line above `name : '_did.'` to be this `name : `_did.${didDocument.id.split(':')[2]}`. i'm not sure where we create auth gateway records, but it would be a similar pattern there.

thehenrytsai: Sounds cleaner! Imma try it!

packages/dids/tests/methods/did-dht.spec.ts
thehenrytsai: Not sure I get what you mean, let's chat!

nitro-neal: Oh just a rename, good to go!

diehuxx (approved):
looks good! +1 to @nitro-neal's comment, I would love to see the new test vectors added to the web5-spec

nitro-neal (approved):

csuwildcat (approved):

PR #507, opened 2024-05-02
add reference types from dwn-sdk-js to avoid pnpm build error

LiranCohen:
`pnpm` encounters some build errors for inferred types when there are 2 different versions of the same package in a monorepo.

https://github.com/microsoft/TypeScript/issues/47663#issuecomment-1962129199

Adding `/// <reference types="@tbd54566975/dwn-sdk-js" />` at the top of files which would infer these types allows the build system to understand them, i believe while keeping this package as a devDependency.

I don't think that this is the long-term solution here, but may be a temporary fix until we decide on a permanent one potentially bringing in `dwn-sdk-js` as a first class dependency of `api` and get rid of the type-wrapping we are doing in `agent`

### DIFF
--- a/.changeset/beige-tools-yawn.md
+++ b/.changeset/beige-tools-yawn.md
@@ -1,0 +1,7 @@
+---
+"@web5/api": patch
+---
+
+add reference types from dwn-sdk-js to avoid pnpm build error
+
+https://github.com/microsoft/TypeScript/issues/47663#issuecomment-1962129199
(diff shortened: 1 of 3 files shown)


shamilovtim (dismissed):

thehenrytsai (approved):
Consider adding an in-line comment to the type reference to explain why it is added, else it might not be easy/obvious to reverse engineer why the line is there.

thehenrytsai (approved):
🙏

PR #513, opened 2024-05-03
Add a helper method for generating a PaginationCursor

LiranCohen:
Currently the only way to get a pagination cursor is from a query response that has more than the number of records you've queried for.

However there could be some cases where the user wants to keep track of the last record they have retrieved and retrieve subsequent records.

Added some helper methods to `agent` in order to build/return the data needed, and a convenience method to the `Record` class in `api` that utilizes them.

### DIFF
--- a/.changeset/odd-eels-rest.md
+++ b/.changeset/odd-eels-rest.md
@@ -1,0 +1,6 @@
+---
+"@web5/agent": patch
+"@web5/api": patch
+---
+
+Add a helper methods for generating a PaginationCursor from `api` without importing `dwn-sdk-js` directly
--- a/.changeset/odd-eels-rest.md
+++ b/.changeset/odd-eels-rest.md
@@ -1,4 +1,7 @@
 ---
+"@web5/identity-agent": patch
+"@web5/proxy-agent": patch
+"@web5/user-agent": patch
 "@web5/agent": patch
 "@web5/api": patch
 ---
--- a/packages/agent/src/utils.ts
+++ b/packages/agent/src/utils.ts
@@ -65,7 +65,7 @@ export async function getPaginationCursor(record: RecordsWriteMessage, dateSort:
     record.descriptor.dateCreated : record.descriptor.datePublished;
 
   if (value === undefined) {
-    throw new Error('The dateCreated or datePublished property is missing from the record descriptor.')
+    throw new Error('The dateCreated or datePublished property is missing from the record descriptor.');
   }
 
   return {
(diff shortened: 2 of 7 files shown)


packages/api/src/record.ts
```diff
@@ -576,6 +579,16 @@ export class Record implements RecordModel {
...
   }
 
+  /**
+   * Returns a pagination cursor for the current record given a sort order.
+   *
+   * @param sort the sort order to use for the pagination cursor.
+   * @returns A promise that resolves to a pagination cursor for the current record.
+   */
+  async paginationCursor(sort: DwnDateSort): Promise<DwnPaginationCursor> {
+    return getPaginationCursor(this.rawMessage, sort);
```
shamilovtim, line 589: I've noticed there's this leaky abstraction in some of the monorepo. Currently, the agent cannot be used independently of the api, which leads me to question the necessity of duplicating methods across both.

The  paginationCursor method in api here merely redirects to getPaginationCursor in agent, without adding any additional logic. I think we should start reducing instances of this and fix the leaky abstractions proactively rather than letting them fester. What do you think?

packages/agent/src/utils.ts
```diff
@@ -55,6 +56,33 @@ export function isRecordsWrite(obj: unknown): obj is RecordsWrite {
   );
 }
 
+/**
+ * Get the CID of the given RecordsWriteMessage.
+ */
+export function getRecordMessageCid(message: RecordsWriteMessage): Promise<string> {
+  return Message.getCid(message);
```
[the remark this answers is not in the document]
LiranCohen: The pattern we have going right now is to not import `dwn-sdk-js` directly into `@web5/api` and instead have `@web5/agent` handle that.

The only instances where we bring `dwn-sdk-js` into `@web5/api` is for testing, and as of last release the type inferences, which is only in `devDependencies`.

I agree that the pattern causes duplication, but I'd rather not break that pattern with this PR and instead spend some time thoroughly discussing it amongst ourselves, at which case there should be a single effort to correct that.

packages/api/src/record.ts
LiranCohen: Mostly same comments as above, however I'm not sure about something.

> Currently, the agent cannot be used independently of the api.

Can you elaborate on that? It should be the other way around, the `api` cannot be used independently of `agent`.

.changeset/odd-eels-rest.md
```diff
@@ -0,0 +1,6 @@
+---
+"@web5/agent": patch
+"@web5/api": patch
```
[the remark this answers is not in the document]
LiranCohen: Technically speaking yes, but because we are pre 1.0 these would be patches since they do not break any existing functionality but just add methods.

packages/api/src/record.ts
shamilovtim: We will revisit after some back and forth with the team

shamilovtim (dismissed):

shamilovtim (approved):

Issue #3, opened 2023-03-08
Subscribe api


michaelneale:
It would make sense in web5 api to have a subscribe api to "listen" for changes that match a filter for (example) your own DID (eg if you are building a messaging system). is this possible?

csuwildcat:
Yes, there's no reason we can't introduce a 'live stream' style method to the Records interface, like `Web5.records.feed()`, that takes a query filter of the same type RecordsQuery already does and establishes a streamed feed of all things that match it. This should be something @mistermoe and @thehenrytsai can weigh in on and add in some form to service the need.

mistermoe:
Code already exists in aggregator. Can easily lift it and drop it into here

csuwildcat:
@mistermoe can you, I, @michaelneale, @thehenrytsai, and @frankhinek do a little session to walk through the requirements for a feed feature? I am wondering how it could manifest in a general way for the API surface.

michaelneale:
yeah lets

michaelneale:
we can chat tomorrow - but the basic idea is that if web5-sdk is connecting to a did that is remote (eg an aggregator style DWN) is there a way we can have a callback type of api for a subscriube that matches a certain filter. 

Use cases: 

real time messaging to remote DWN
Dropbox type of thing
TBDex
Avoiding aggro-polling some remote server if you want to sync it to local unaddressable node 

etc

did I capture that @mistermoe?

something like:


```
const response = await Web5.records.subscribe('did:example:bob', {
  author: 'did:example:alice',
  message: {
    filter: {
      schema: 'https://schema.org/Playlist',
      dataFormat: 'application/json'
    }
  }
}, function(result) {  .... } );
```

Issue #497, opened 2024-04-23
Update did:dht to conform to the latest version of the specification


decentralgabe:
There have been a number of changes to the DID DHT spec since implementation, which I've captured below. These changes should be backwards compatible, meaning DIDs you use today will not break.

These changes are already in place in the [DID DHT project itself](https://github.com/TBD54566975/did-dht-method/blob/main/impl/internal/did/did.go), and you can find [passing test vectors here](https://github.com/TBD54566975/did-dht-method/blob/main/impl/internal/did/did_test.go#L228) as a reference.

1. Support the three test vectors listed here: https://did-dht.com/#test-vectors
2. Support NS records to be set to the publishing gateway (as per https://did-dht.com/#designating-authoritative-gateways, https://github.com/TBD54566975/did-dht-method/issues/129)
3. Always set alg and kid on expansion (to the values in the registry), support overriding of alg values (as per https://github.com/TBD54566975/did-dht-method/pull/148)
4.  Support TXT records over 255 characters (as per https://github.com/TBD54566975/did-dht-method/pull/162)
5. Set verification method's controller property, allow overriding (as per https://github.com/TBD54566975/did-dht-method/pull/102)
6. Stamp version field in the root record - set to 1 for now (as per https://github.com/TBD54566975/did-dht-method/pull/102)
7. Support multiple service endpoints (as per https://github.com/TBD54566975/did-dht-method/pull/86)
8. Support X25519 (as per https://github.com/TBD54566975/did-dht-method/pull/148) - note this is required for the 3rd test vector

Additionally, I would recommend:

9. Support a `republish` feature, which is able to take the latest state of a DID DHT document and re-publish it to a gateway. 

If not in practice already, it would make sense to expand our "PortableDID" construct to store histories of DID DHT documents and republish should only work on the latest version (seq num).

----------------------------------------------

After this you can anticipate two additional changes to DID DHT pertaining to the following issues:
- https://github.com/TBD54566975/did-dht-method/issues/74
- https://github.com/TBD54566975/did-dht-method/issues/100

74 will be a breaking change, though (9) from above paves a path to make it less disruptive.

Please let me know if you have any questions.

nitro-neal:
ok so in web5-js 

1. Support the three test vectors listed here ❌  (Needs to be fully implemented)
2. Support NS records to be set to the publishing gateway  ❌  (needs to be fully implemented)
3. Always set alg and kid on expansion 🟡 ❓  (not sure if this exists currently)
[...]

PR #518, opened 2024-05-07
Always set alg and kid on expansion

nitro-neal:
This pr resolves https://github.com/TBD54566975/web5-js/issues/3 from this issue:
https://github.com/TBD54566975/web5-js/issues/497

Always set alg and kid on expansion (to the values in the registry), support overriding of alg values

toDnsPacket:
* check if the alg on the JWK == the alg in the default registry for the given key type. Otherwise add it to the txt document as `a`

fromDnsPacket:
* KID is always either 0 for the identity key or the JWK thumbprint for all other keys (no overriding it)
* the alg is always set to the default from the registry (https://did-dht.com/registry/#key-type-index) but can be overriden if you specify the alg property

### DIFF
--- a/packages/dids/src/methods/did-dht.ts
+++ b/packages/dids/src/methods/did-dht.ts
@@ -32,7 +32,6 @@ import { extractDidFragment } from '../utils.js';
 import { DidError, DidErrorCode } from '../did-error.js';
 import { DidVerificationRelationship } from '../types/did-core.js';
 import { EMPTY_DID_RESOLUTION_RESULT } from '../types/did-resolution.js';
-import { getJoseSignatureAlgorithmFromPublicKey } from '@web5/crypto/utils';
 
 /**
  * Represents a BEP44 message, which is used for storing and retrieving data in the Mainline DHT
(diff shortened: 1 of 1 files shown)


decentralgabe (dismissed):

packages/dids/src/methods/did-dht.ts
```diff
@@ -1026,6 +1036,12 @@ export class DidDhtDocument {
           // Convert the public key from a byte array to JWK format.
           let publicKey = await DidDhtUtils.keyConverter(namedCurve).bytesToPublicKey({ publicKeyBytes });
 
+          // DID DHT spec requires `alg` in keys in the DID document
+          publicKey.alg = parsedAlg || getJoseSignatureAlgorithmFromPublicKey(publicKey);
```
thehenrytsai, line 1040: `getJoseSignatureAlgorithmFromPublicKey` seems to return `EdDSA` when `crv` is `Ed25519`.

Maybe what need to be done is to extend the vector tests such that:

DID document -> DNS records -> DID document ends up being the same. Right now the vector tests stops at DNS records conversion.

thehenrytsai (approved):
🚀 🚀 🚀

PR #566, opened 2024-05-13
Add new Web features module with networking feature installation

csuwildcat:


nitro-neal:
I would double check this, seems like all the dependabot stuff would be merged in too

LiranCohen:
Features already implemented in a different branch that was merged.

PR #576, opened 2024-05-16
Add ns record gateway handling

nitro-neal:
When resolving a did dht this change allows it to look at the ns records and find the authoritative gateways.

It will try to use these gateways, and accumulate errors if it fails. It will keep trying until all gateway uris are exhausted and then will return the accumulated errors.

1. Do a resolution with default gateway to get dns records for did
2. If there are NS records, get all of them into an array of endpoints
3. Loop over all endpoints, if you successfully get dns records for did from endpoint, return and continue to resolve did, otherwise try the next endpoint

Reference
https://did-dht.com/#resolution-gateways

### DIFF
--- a/packages/dids/src/methods/did-dht.ts
+++ b/packages/dids/src/methods/did-dht.ts
@@ -1020,7 +1020,6 @@ export class DidDhtDocument {
       if (answer.name.endsWith(`.${DidDhtDocument.getUniqueDidSuffix(didUri)}.`)) {
         const gatewayUri = answer.data.slice(0, -1); // Remove trailing dot
         authoritativeGatewayUris.push(gatewayUri);
-        break;
       }
     }
 
--- a/packages/dids/tests/methods/did-dht.spec.ts
+++ b/packages/dids/tests/methods/did-dht.spec.ts
@@ -1159,70 +1159,75 @@ describe('DidDhtDocument', () => {
     });
 
     it('handles custom authoritative gateways', async () => {
+
+      const didDocument: DidDocument = {
+        id                 : 'did:dht:5cahcfh3zh8bqd5cn3y6inoea1b3d6kh85rjksne9e5dcyrc1ery',
+        verificationMethod : [
+          {
+            id           : 'did:dht:5cahcfh3zh8bqd5cn3y6inoea1b3d6kh85rjksne9e5dcyrc1ery#0',
+            type         : 'JsonWebKey',
+            controller   : 'did:dht:5cahcfh3zh8bqd5cn3y6inoea1b3d6kh85rjksne9e5dcyrc1ery',
+            publicKeyJwk : {
+              crv : 'Ed25519',
+              kty : 'OKP',
+              x   : '2zHGF5m_DhcPbBZB6ooIxIOR-Vw-yJVYSPo2NgCMkgg',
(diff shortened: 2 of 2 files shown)


packages/dids/src/methods/did-dht.ts
```diff
@@ -805,10 +805,33 @@ export class DidDhtDocument {
...
     // Retrieve the signed BEP44 message from a DID DHT Gateway or Pkarr relay.
-    const bep44Message = await DidDhtDocument.pkarrGet({ gatewayUri, publicKeyBytes });
+    let bep44Message = await DidDhtDocument.pkarrGet({ gatewayUri, publicKeyBytes });
 
     // Verify the signature of the BEP44 message and parse the value to a DNS packet.
-    const dnsPacket = await DidDhtUtils.parseBep44GetMessage({ bep44Message });
+    let dnsPacket = await DidDhtUtils.parseBep44GetMessage({ bep44Message });
+
+    // Look at the NS records in the DNS packet to find the resolution gateway URIs.
+    let resolutinGatewayUris = await DidDhtDocument.getAuthoritativeGatewayUris({ didUri, dnsPacket });
```
decentralgabe, line 814:
```suggestion
    let resolutionGatewayUris = await DidDhtDocument.getAuthoritativeGatewayUris({ didUri, dnsPacket });
```

packages/dids/src/methods/did-dht.ts
```diff
@@ -805,10 +805,36 @@ export class DidDhtDocument {
...
 
     // Verify the signature of the BEP44 message and parse the value to a DNS packet.
-    const dnsPacket = await DidDhtUtils.parseBep44GetMessage({ bep44Message });
+    let dnsPacket = await DidDhtUtils.parseBep44GetMessage({ bep44Message });
+
+    // Look at the NS records in the DNS packet to find the resolution gateway URIs.
+    let resolutionGatewayUris = await DidDhtDocument.getAuthoritativeGatewayUris({ didUri, dnsPacket });
+
+    // Only do a second retrieval if the authoritative resolution gateway URIs are different from the given gateway URI.
+    if(!resolutionGatewayUris.includes(gatewayUri)) {
```
decentralgabe, line 817: does this work with the `.` / FQDN syntax?

decentralgabe (approved):
approving if the condition I mentioned as test coverage

packages/dids/src/methods/did-dht.ts
thehenrytsai, line 817: This is confusing, the intent seems to only enter the condition check if there is gateway URI specified and the given URI is not in that list. But right now the code inside is still executed when the array is empty.

thehenrytsai: I have a small refactoring PR that address this:

https://github.com/TBD54566975/web5-js/pull/608

thehenrytsai (commented):
Looks legit, only missing test for the code added.

Also added a small refactoring PR to merge that addresses a few small issues: https://github.com/TBD54566975/web5-js/pull/608

packages/dids/src/methods/did-dht.ts
nitro-neal: thanks for catching! I merged it into this branch

nitro-neal:
I will wait to merge until vector 3 is in, and then I will resolve conflict and finish testing missing lines

PR #633, opened 2024-05-21
Updated @types/eslint dependencies across all packages

thehenrytsai:
1. Updated @types/eslint dependencies across all packages. They need to be updated together else `tsc` will result in `error TS2300: Duplicate identifier` errors.
1. Opportunistically updated `playwright` dependencies.

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -78,7 +78,7 @@
     "@playwright/test": "1.44.0",
     "@types/chai": "4.3.16",
     "@types/chai-as-promised": "7.1.8",
-    "@types/eslint": "8.44.2",
+    "@types/eslint": "8.56.10",
     "@types/mocha": "10.0.6",
     "@types/node": "20.12.12",
     "@types/readable-stream": "4.0.9",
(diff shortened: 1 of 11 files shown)


packages/common/package.json
```diff
@@ -95,7 +95,7 @@
     "eslint-plugin-mocha": "10.4.3",
     "mocha": "10.4.0",
     "mocha-junit-reporter": "2.2.1",
-    "playwright": "1.40.1",
+    "playwright": "1.44.0",
```
nitro-neal, line 98: did this get updated in other places?

nitro-neal: I notice its a different version in packages/agent/package.json

nitro-neal (approved):

packages/common/package.json
thehenrytsai: I did not update agent related packages because they are not part of the dependabot check.

LiranCohen (approved):

csuwildcat (approved):

PR #635, opened 2024-05-22
Implement a `delete` method and state on the `@web5/api` `Record` class

LiranCohen:
This PR Allows a `Record` class to include a `deleted` state based on a `RecordsDeleteMessage` being applied to it.

We update the `RecordModel` to reflect the immutable properties that a record may contain, as well as the optional mutable properties that are only relevant to a record that is in a non-deleted state.

When a record is in a deleted state, it derives the immutable properties from the `initialWrite` of that record.

Currently when dealing with a remote DWN, the user cannot be notified of a deleted state through a Query, however when Subscriptions are brought up to the API a user could receive a message from the remote which signals that the record has been deleted, more useful scenario testing will be implemented when `record.subscribe()` feature is added.

- Implements a `delete()` method on the `Record` class that behaves similarly to `update()`.
- Implements a `deleted` property on the `Record` class to indicate if the record is in a deleted state.

### DIFF
--- a/packages/api/tests/record.spec.ts
+++ b/packages/api/tests/record.spec.ts
@@ -2077,7 +2077,7 @@ describe('Record', () => {
       expect(recordString).to.contain(`ID: ${record.id}`);
       expect(recordString).to.contain(`Context ID: ${record.contextId}`);
       expect(recordString).to.contain(`Protocol: ${record.protocol}`);
-      expect(recordString).to.contain(`Schema : ${record.schema}`);
+      expect(recordString).to.contain(`Schema: ${record.schema}`);
       expect(recordString).to.contain(`Deleted: ${false}`); // record is not deleted
       expect(recordString).to.contain(`Created: ${record.dateCreated}`);
       expect(recordString).to.contain(`Modified: ${record.dateModified}`);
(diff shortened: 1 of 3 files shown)


packages/api/tests/record.spec.ts
```diff
@@ -2084,6 +2167,56 @@ describe('Record', () => {
       expect(updatedData).to.equal('bye');
     });
 
+    it('updates a record to be unpublished from published', async () => {
```
LiranCohen, line 2170: Added test to increase missing coverage

frankhinek:
@LiranCohen worth noting that this exact functionality was removed in a prior PR:

https://github.com/TBD54566975/web5-js/pull/327

Relevant section:

> Removes the record.delete() instance method from Record objects. This method is the only one that is a duplicate of the identically named web5.dwn.records.* method and introduces unintuitive and error prone behavior when the record was initially retrieved from a remote DWN. Another class of issues is then introduced depending on whether that record already exists locally. In order to remove this "foot gun" and keep the DWN Records API provided by @web5/api free of duplicates that behavior differently depending on unintuitive context, the instance method was removed.

It's been a while but @csuwildcat, @angiejones , and @blackgirlbytes might recall some other context of the motivation for removing this.

LiranCohen:
RE the previous "foot gun", we now account for the `initialWrite` within the Record class and can store/import the state of a deleted record even if it did not already exist locally.

I was thinking in a subsequent PR to remove the `web5.dwn.records.delete()` method as I figured that would warrant it's own discussion. But since it was brought up maybe the discussion can happen here (and functionality can still be removed in a subsequent PR to keep things smaller and reviews easier).

Are there any cases where you see it valid to be able to issue a delete while ONLY knowing the `recordId` of the record you're looking to delete without having that record class already on hand? If that is the case maybe we do leave both (although I agree there is some confusion there). Def worth a discussion.

Worth noting that this functionality is useful for the upcoming `dwn.records.subscribe()` method, as you could be subscribed to a set of records and receive the deleted state for a record you have previously received a write/update state for.

Additionally there have been discussions of allowing `query()` to optionally return deleted states, where this would also become useful.

frankhinek (approved):
🚀 Ship it!

thehenrytsai (approved):
Rubber stamp. 🚀

csuwildcat (approved):

PR #641, opened 2024-05-24
Normalize Alg To EdDSA

nitro-neal:
This change addresses this issue:
When a JWT is created, the alg is set to EdDSA and if not specified, the DID DHT implementation will insert an alg of Ed25519 if the kty is Ed25519. JWT Verify checks whether the header and public key JWK alg values match, and if they don't verification fails.

This change allows for Ed25519 and EdDSA to be considered equivalent when it comes to JWT verification.

This change also adds a version bump patch:

## Summary:
This patch updates the JWT verification logic for did:dht and adds support for status list credentials.

### JWT Verification for did:dht:

* Updated the algorithm (alg) check to ensure that the alg value in the JWT header matches the normalized alg value in the public key JWK.

### Status List Credentials:

`StatusListCredential` represents a digitally verifiable status list credential according to the
[W3C Verifiable Credentials Status List v2021](https://www.w3.org/community/reports/credentials/CG-FINAL-vc-status-list-2021-20230102/).

* Added support for creation of Status List Credentials
* Added support for verifying if a credential is revoked in it's corresponding status list credential.


packages/credentials/tests/verifiable-credential.spec.ts
```diff
@@ -185,6 +185,40 @@ describe('Verifiable Credential Tests', () => {
...
+        issuer  : did.uri,
+        data    : {
+          username: 'nitro'
+        }
+      });
+
+      const vcJwt = await vc.sign({ did });
+
+      await VerifiableCredential.verify({ vcJwt });
+
```
shamilovtim, line 212:
```suggestion
expect(vcJwt.header.alg).to.equal('EdDSA');
```

nitro-neal:
This PR may not be needed anymore since this pr updates the did dht spec - https://github.com/TBD54566975/web5-js/pull/653

thehenrytsai:
No longer needed because we change DID DHT spec.

PR #646, opened 2024-05-29
fix space between @ and handle in CODEOWNERS file

LiranCohen:


thehenrytsai (approved):
😮 
Could this be the reason for strange behaviors...?! I certainly hope so!!

Issue #638, opened 2024-05-24
Allow KeyManager `generateKey()` to accept a custom `kid`


LiranCohen:
We currently do not provide the means for which to allow a caller to specify a custom `kid` within the KeyManager's `generateKey` interface method.

This is particularly needed in order to achieve compliance with the `did:dht` spec, which requires that Verification Method IDs match the JWK's `kid` field.

- [ ] Allow the caller of `generateKey` to pass a custom `kid`
- [ ] Modify the `did:dht` implementation to respect Verification Method IDs to match the JWK's `kid` field.

PR #653, opened 2024-05-31
Update DID DHT Key Type to EdDSA

nitro-neal:
This bug fixes this issue and changes the index 0 key type from `Ed25519` to `EdDSA` (pending change coming to did-dht.com)

image
 
Here was the issue that was happening which this  pr fixes:

When creating a portable did with a jwk like this:
```
        privateKeys: [
          {
            crv : 'Ed25519',
            d   : 'hdSIwbQwVD-fNOVEgt-k3mMl44Ip1iPi58Ex6VDGxqY',
            kty : 'OKP',
            x   : 'VYKm2SCIV9Vz3BRy-v5R9GHz3EOJCPvZ1_gP1e3XiB0',
            kid : 'cyvOypa6k-4ffsRWcza37s5XVOh1kO9ICUeo1ZxHVM8',
            alg : 'EdDSA',
          }
        ]
```
        
(notice Alg is EdDSA)

the resolver gives:
```
    "verificationMethod": [
      {
        "id": "did:dht:ksbkpsjytbm7kh6hnt3xi91t6to98zndtrrxzsqz9y87m5qztyqo#0#0",
        "type": "JsonWebKey",
        "controller": "did:dht:ksbkpsjytbm7kh6hnt3xi91t6to98zndtrrxzsqz9y87m5qztyqo#0",
        "publicKeyJwk": {
          "kty": "OKP",
          "crv": "Ed25519",
          "x": "VYKm2SCIV9Vz3BRy-v5R9GHz3EOJCPvZ1_gP1e3XiB0",
          "kid": "cyvOypa6k-4ffsRWcza37s5XVOh1kO9ICUeo1ZxHVM8",
          "alg": "Ed25519"
        }
      }
    ],
```

(notice alg is Ed25519)

The solution was to change the actual did dht spec so that this does not happen and allows use to verify jwts with did dht

### DIFF
--- a/packages/credentials/tests/jwt.spec.ts
+++ b/packages/credentials/tests/jwt.spec.ts
@@ -3,7 +3,7 @@ import type { JwtHeaderParams, JwtPayload, PrivateKeyJwk } from '@web5/crypto';
 import { expect } from 'chai';
 import { Convert } from '@web5/common';
 import { Ed25519 } from '@web5/crypto';
-import { DidDht, DidJwk, DidKey, PortableDid, UniversalResolver } from '@web5/dids';
+import { DidDht, DidJwk, DidKey, PortableDid } from '@web5/dids';
 
 import { Jwt } from '../src/jwt.js';
 import JwtVerifyTestVector from '../../../web5-spec/test-vectors/vc_jwt/verify.json' assert { type: 'json' };
@@ -169,17 +169,17 @@ describe('Jwt', () => {
 
       const siopv2Response = {
         id_token: await Jwt.sign({
-          signerDid: bearerDid,
-          payload: {
-            iss: bearerDid.uri,
-            sub: bearerDid.uri,
-            aud: 'did:dht:ho3axp5pgp4k8a7kqtb8knn5uaqwy9ghkm98wrytnh67bsn7ezry',
-            nonce: 'd844f80d21c33ea6e087afa2b84dc31f',
-            iat: Math.floor(Date.now() / 1000),
-            exp: Math.floor(Date.now() / 1000) + (30 * 60), // plus 30 minutes
+          signerDid : bearerDid,
+          payload   : {
+            iss   : bearerDid.uri,
+            sub   : bearerDid.uri,
+            aud   : 'did:dht:ho3axp5pgp4k8a7kqtb8knn5uaqwy9ghkm98wrytnh67bsn7ezry',
+            nonce : 'd844f80d21c33ea6e087afa2b84dc31f',
+            iat   : Math.floor(Date.now() / 1000),
+            exp   : Math.floor(Date.now() / 1000) + (30 * 60), // plus 30 minutes
           }
         })
-      }
+      };
 
       // Verify the JWT and make sure we get a result that it does not throw an error.
       const jwtVerifyResult = await Jwt.verify({ jwt: siopv2Response.id_token });
--- a/packages/credentials/tests/jwt.spec.ts
+++ b/packages/credentials/tests/jwt.spec.ts
@@ -73,17 +73,8 @@ describe('Jwt', () => {
   });
 
   describe('verify()', () => {
-    let dereferenceStub: sinon.SinonStub;
-
-    beforeEach(() => {
-      dereferenceStub = sinon.stub(Jwt.didResolver, 'dereference');
-    });
-
-    afterEach(() => {
-      dereferenceStub.restore();
-    });
-
     it('successful verify with did:dht', async () => {
+      const dereferenceStub = sinon.stub(Jwt.didResolver, 'dereference');
 
       const mockResult: DidDereferencingResult = {
         dereferencingMetadata: {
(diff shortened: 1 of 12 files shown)


pnpm-lock.yaml
```diff
@@ -330,14 +330,14 @@ importers:
         specifier: 2.1.0
         version: 2.1.0
       '@web5/common':
-        specifier: 1.0.0
-        version: 1.0.0
+        specifier: 1.0.1
+        version: link:../common
```
nitro-neal, line 334: is this fine?

packages/credentials/tests/jwt.spec.ts
```diff
@@ -72,6 +82,113 @@ describe('Jwt', () => {
...
+      fetchStub.resolves(fetchOkResponse());
+    });
+
+    afterEach(() => {
+      fetchStub.restore();
+    });
+
+    it('successful verify with did:dht', async () => {
+      const hexString =
+        '0ab2b3386e22595e1271e7ef67fda70c37acf7d28b8c884a6fdcbb0ea739f341' +
```
thehenrytsai, line 102: Not a fan of opaque inputs to tests. Can you update and enable vector tests accordingly for the js implementation and call out in https://github.com/TBD54566975/web5-js/issues/638  that the kid changes must be reverted when 638 is fixed.

I think they should cover this change. Let me know if I am not making sense!

nitro-neal:
I have turned the test vectors back on, they can pass by either:
1. what i've done here in the pr, changing the kid in the test vector
OR 
2. I can add this line to our fromDnsPacket parsing:
```js
          if(dnsRecordId === 'k0') {
            publicKey.kid = '0';
          }
```

image


Not sure which way is right at this point! but feel free to let me know which way we want to go

packages/credentials/tests/jwt.spec.ts
nitro-neal: This test tests the Jwt.sign / verify with did dht, (the reason we had to change the spec a bit in the first place)

decentralgabe (dismissed):

packages/dids/tests/fixtures/test-vectors/did-dht/vector-2.json
```diff
@@ -9,8 +9,8 @@
         "type": "JsonWebKey",
         "controller": "did:dht:cyuoqaf7itop8ohww4yn5ojg13qaq83r9zihgqntc5i9zwrfdfoo",
         "publicKeyJwk": {
-          "kid": "0",
-          "alg": "Ed25519",
+          "kid": "HP6bkG6mv-YPsU3Vi5Coi5wGYSVW9abFBwBSQwLQ7PU",
```
decentralgabe, line 12:
```suggestion
          "kid": "0",
```

packages/dids/tests/fixtures/test-vectors/did-dht/vector-1.json
```diff
@@ -7,8 +7,8 @@
         "type": "JsonWebKey",
         "controller": "did:dht:cyuoqaf7itop8ohww4yn5ojg13qaq83r9zihgqntc5i9zwrfdfoo",
         "publicKeyJwk": {
-          "kid": "0",
-          "alg": "Ed25519",
+          "kid": "HP6bkG6mv-YPsU3Vi5Coi5wGYSVW9abFBwBSQwLQ7PU",
```
decentralgabe, line 10:
```suggestion
          "kid": "0",
```

nitro-neal:
I have updated, 

added this line:
```
   if(dnsRecordId === 'k0') {
     publicKey.kid = '0';
   }
```

And chnaged the test vectors to have kid as 0 again

packages/dids/tests/methods/did-dht.spec.ts
```diff
@@ -1298,7 +1298,7 @@ describe('Official DID:DHT Vector tests', () => {
   // Temporarily disabling due to inability to add a custom `kid` to our JWK in the current deployment.
   // We disable the test instead of changing the official specd test vector.
   // TODO: https://github.com/TBD54566975/web5-js/issues/638
```
thehenrytsai, line 1300: Get rid of these, no longer applicable.

packages/dids/src/methods/did-dht.ts
```diff
@@ -1055,6 +1055,10 @@ export class DidDhtDocument {
 
           publicKey.alg = parsedAlg || KeyTypeToDefaultAlgorithmMap[Number(t) as DidDhtRegisteredKeyType];
 
+          if(dnsRecordId === 'k0') {
+            publicKey.kid = '0';
+          }
+
```
thehenrytsai, line 1061: Can't do this, else will break things. This is what @LiranCohen reverted. You should update the vector instead to use the thumbprint as `kid` temporarily. And optionally update #638 to mention what you changed so @LiranCohen can revert when #638 is done.

packages/credentials/tests/jwt.spec.ts
thehenrytsai: Got it, assuming the main intent of this test is to test sign and verify, can we avoid using the opaque hex string in the test since it is besides the point?

packages/dids/src/methods/did-dht.ts
nitro-neal: updated

thehenrytsai (approved):
🚀

PR #691, opened 2024-06-13
Add sw cache and link features

csuwildcat:


shamilovtim (commented):
Is there a way we're excluding this DOM code from the bundle when it's consumed by other environments? 

Since there's nothing inherently browser oriented about the `/api` package. I wonder if this should be living in `user-agent` which is kind of the package for browser clients?

LiranCohen (dismissed):

csuwildcat:
Looks like an unrelated VC test is failing for timeout reasons, probably due to resolution? We should be able to proceed anyway.

LiranCohen (approved):
🚀

PR #693, opened 2024-06-21
Expose Web5PlatformAgent to Web5 instance

bnonni:
We need an easier way to access the keys of the connectedDid in the Web5.connect response object. This PR exposes the `Web5PlatformAgent` directly in the Web5 object. Example use case we have in our current app:

```ts
const { web5, did: aliceDid, recoveryPhrase } = await Web5.connect({ password });
const { did: aliceBearerDid } = await web5.platformAgent.identity.get({ didUri: aliceDid });
const signer = await aliceBearerDid.getSigner();
const signature = await signer.sign({ data: 'Hello world!' });
```

### DIFF
--- a/packages/api/src/web5.ts
+++ b/packages/api/src/web5.ts
@@ -1,4 +1,4 @@
-import type { BearerIdentity, HdIdentityVault, Web5Agent } from '@web5/agent';
+import type { BearerIdentity, HdIdentityVault, Web5Agent, Web5PlatformAgent } from '@web5/agent';
 
 import { Web5UserAgent } from '@web5/user-agent';
 
@@ -143,6 +143,20 @@ export class Web5 {
   /** Exposed instance to the VC APIs, allow users to issue, present and verify VCs */
   vc: VcApi;
 
+  /**
+   * A {@link Web5PlatformAgent} instance exposing `identity` {@link AgentIdentityApi} to the Web5
+   * instance to provide direct access to the signing ability of the connectedDid.
+   * 
+   * @example 
+   * ```ts
+   * const { web5, did: aliceDid, recoveryPhrase } = await Web5.connect({ password });
+   * const { did: aliceBearerDid } = await web5.platformAgent.identity.get({ didUri: aliceDid });
+   * const signer = await aliceBearerDid.getSigner();
+   * const signature = await signer.sign({ data: 'Hello world!' });
+   * 
+   */
+   platformAgent: Web5PlatformAgent;
+
   /** The DID of the tenant under which DID operations are being performed. */
   private connectedDid: string;
 
@@ -152,6 +166,7 @@ export class Web5 {
     this.did = new DidApi({ agent, connectedDid });
     this.dwn = new DwnApi({ agent, connectedDid });
     this.vc = new VcApi({ agent, connectedDid });
+    this.platformAgent = agent as Web5PlatformAgent;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


packages/api/src/web5.ts
```diff
@@ -152,6 +166,7 @@ export class Web5 {
     this.did = new DidApi({ agent, connectedDid });
     this.dwn = new DwnApi({ agent, connectedDid });
     this.vc = new VcApi({ agent, connectedDid });
+    this.platformAgent = agent as Web5PlatformAgent;
```
shamilovtim, line 169: Since `Web5PlatformAgent` is purely a compile time Type that gets stripped out in production, I don't think any runtime code is necessary to solve for what's being solved for here

shamilovtim:
You should be able to immediately fix this in your downstream application by asserting that your `agent` is of type `Web5PlatformAgent` (`agent as Web5PlatformAgent`) without any upstream changes. But an upstream change that might immediately fix this is a type guard (or some sort of generic) to assert that an upstream agent is of type `Web5PlatformAgent`

frankhinek:
@bnonni is the app you are including this in a client/user facing app?  or is it a backend service?

packages/api/src/web5.ts
frankhinek: Yea agreed.  I don't think this is a pattern we'd want to follow and have duplicate properties for the same thing.

bnonni:
> You should be able to immediately fix this in your downstream application by asserting that your `agent` is of type `Web5PlatformAgent` (`agent as Web5PlatformAgent`) without any upstream changes. But an upstream change that might work for this PR is either a type guard or a generic

@shamilovtim I don't know why I did not think of this 🤦🏼‍♂️ I must be tired 😪 This works, thank you for the suggestion.

bnonni:
> @bnonni Is the app you are developing a web browser client/user facing app? Or is it a backend service or mobile app?
> 
> If its a mobile app or backend service, then you probably don't need to use `Web5.connect()` since you're never going to be connecting to any remote agents.
> 

@frankhinek both server and client. [DCX](https://github.com/TBD54566975/incubation-dcx) will be using `Web5.connect` on the server side as a [DCX issuer](https://github.com/TBD54566975/incubation-dcx/blob/main/src/protocol/credential-issuer.ts) and on the client side as a [DCX applicant](https://github.com/TBD54566975/incubation-dcx/blob/main/src/protocol/credential-applicant.ts). We have both a mobile and web app client side, so the web version would need `Web5.connect`. Each side needs access to their respective DWNs to perform their side of the DCX protocol.

> Also, if this need to use the underlying agent is related to a lack of `web5.vc.create()` then we likely ought to tackle adding that in a PR rather than adding a second agent property to the `Web5` instance.

Yes, that's accurate. Having an easy method to issue a vc from the server side is what I'm looking to do. To your point, I will close this PR and open one for the `web5.vc.create()` functionality instead.

bnonni:
@shamilovtim @frankhinek 

I don't fully understand the nuances of the agent did vs. connected DID, and I'm concerned that when we scale DCX into the cloud and replicate it across multiple instances each running `Web5.connect`, they will each have the same agent DID with a different connected DID leading to issues with VC signatures and verification.

- Are the connected DIDs published with their own DID docs?
- Is there a way to resolve a connected DID back to its agent DID?
- Which DID should we use to sign VCs on the server side?
- If we use the connected DID, will the signatures on the VCs we issue to holders be valid against the agent DID somehow?
- I'm aware of the long term plan to have delegation grants between agent DID and connected DIDs. Until that is complete, how should we go about signing and issuing VCs from the server? How do we let clients know a connected DID has a relationship to an agent DID such that the connected DID signature can be verified against the agent DID?

Thanks for the help!

packages/api/src/web5.ts
bnonni: ACK

PR #694, opened 2024-06-21
Added wallet connect options for Web5.connect()

thehenrytsai:
1. Added WalletConnectOptions for the connect flow.
1. Updated to latest `dwn-sdk-js` (v0.3.10)
1. Fixed npm audit vulnerability https://github.com/advisories/GHSA-3h5v-q93c-6h6q
1. Updated "agent" dependencies using `workspace:*` syntax to always keep "agent" dependency in sync to avoid manual updates/mistakes.


packages/agent/src/sync-engine-level.ts
```diff
@@ -355,13 +355,14 @@ export class SyncEngineLevel implements SyncEngine {
...
+    let eventsReply = {} as EventsQueryReply;
 
     if (syncDirection === 'pull') {
       // When sync is a pull, get the event log from the remote DWN.
       const eventsGetMessage = await this.agent.dwn.createMessage({
         author        : did,
-        messageType   : DwnInterface.EventsGet,
+        messageType   : DwnInterface.EventsQuery,
+        // TODO: Liran help!
         messageParams : { cursor }
```
thehenrytsai, line 366: Hey @LiranCohen, `messageParams` current broke, need your help instead of me wasting time fixing, I know you know exactly what needs to be done.

You can directly checkout and merge in fixes in this branch!

thehenrytsai: Thanks @LiranCohen, you are the best.

packages/api/src/web5.ts
```diff
@@ -1,20 +1,76 @@
...
+   */
+  permissionScopes: DwnRecordsPermissionScope[];
+
+  /**
+   * A handler to be called when the request URL is ready to be used to fetch the permission request by the wallet.
+   * This method should be used by the calling app to pass the request URL to the wallet via a QR code or a deep link.
+   *
+   * @param requestUrl - The request URL for the wallet to fetch the permission request.
+   */
+  onRequestReady: (requestUrl: string) => void;
```
shamilovtim, line 56: I don't think this param is necessary

packages/api/src/web5.ts
```diff
@@ -1,20 +1,76 @@
...
+
+  /**
+   * An async method to get the PIN needed from the user to decrypt the response from the wallet.
+   *
+   * @returns A promise that resolves to the PIN as a string.
+   */
+  pinGetter: () => Promise<string>;
```
shamilovtim, line 63: Also don't think this param is necessary

packages/api/package.json
```diff
@@ -77,23 +77,23 @@
     "node": ">=18.0.0"
   },
   "dependencies": {
-    "@web5/agent": "0.3.8",
+    "@web5/agent": "workspace:*",
```
shamilovtim, line 80: intentional? probably needs a revert

shamilovtim (commented):
I think a few changes needed

LiranCohen (approved):
Looks good, but should we include `changesets` with this PR?

Also remember to create a separate one for `api` vs the others.

packages/api/src/web5.ts
thehenrytsai: @shamilovtim, can you elaborate why not?

packages/api/src/web5.ts
thehenrytsai: @shamilovtim, can you elaborate why not? Thanks.

packages/api/package.json
thehenrytsai: @shamilovtim, can you elaborate why this should be reverted? I did this intentionally with the hope that we don't need to keep having two phase version update of doing `agent` followed by the `api`.

shamilovtim: confused: this is an active package that gets published to npm. isn't this an invalid package.json.
isn't it the case that you can't ship this change to end users? also, what about changesets?

thehenrytsai: pnpm will take care of translating the `workspace:*` into the right version.

shamilovtim: i'm not following how this is any easier than predictable versioning that we can reason about with a historical publish record that matches version control

packages/api/package.json
```diff
@@ -81,7 +81,7 @@
     "@web5/common": "1.0.0",
     "@web5/crypto": "1.0.0",
     "@web5/dids": "1.1.0",
-    "@web5/user-agent": "0.3.8"
+    "@web5/user-agent": "workspace:*"
```
shamilovtim, line 84: maybe it's just me but i find this impossible to reason about or manage. are we going to rely on lockfiles or hashes to know what versions went together? 

even when i look at the lockfile, the changed version isn't even tracked and there isn't even a hash.

thehenrytsai: For local development: it will always point to the code on the local machine. The intent is that if a dependent package has a breaking change to downstream mono repo, CICD will not let the PR through.

For released package, you can checkout the commit of  `Version Packages` (either via command or GitHub UI), and see the version.

LiranCohen (dismissed):

packages/api/src/web5.ts
shamilovtim: Actually I see what you're going for here, nevermind

packages/api/package.json
thehenrytsai: The intent is that if a dependent package introduces a breaking change to downstream mono repo, CICD will not let the PR through. Nor would one need to remember to hunt and update downstream packages to depend on the new version.

packages/api/package.json
thehenrytsai: For end-user importing a package or release, they'd know the dependency version the same way they alway would, if they care to know.

shamilovtim: > For local development: it will always point to the code on the local machine

right, that's not the problem with this though.

> For released package, you can checkout the commit of Version Packages (either via command or GitHub UI), and see the dependency versions.

Right, that's the problem getting at is that there will be no versioning of packages under this paradigm. You are sidestepping that flow here. The versions are not being checked into version control. You are opaquely publishing them to npm without accounting for them in version control and planning on pnpm to "take care of translating the workspace:* into the right version." Which does not get checked into version control, as evidenced by this PR. 

Maybe to clarify: is the intention for this to be a permanent change?

packages/api/src/web5.ts
shamilovtim: Just had some naming confusion all good

packages/api/package.json
thehenrytsai: > is the intention for this to be a permanent change?

Yes, I intended for this to be a permanent change. To be fair, I am a complete noob to `pnpm` that was introduced to `web5-js` repo, but the desire for wanting a mono repro is clear to me because the exact same problems existed in previous projects I worked on.

When `npm` publishes package A the depends B, it looks into A's `pacakge.json` for version of B to use.
When `pnpm` publishes package A the depends B, it looks into A's `package.json` for version of B to use, which points to the `package.json` of B that contains the version number.

It's merely difference between how two different tools publishes a package, though sure, `npm` is more "first class", but here we are choosing `pnpm` to augment the limitations of `npm`.

shamilovtim: I get all of that already and it's not the root of the problem I'm raising. Let me explain more clearly maybe. We have a CI flow where we release off of `main` using changesets. Each PR increments the appropriate version as long as you create a changeset. This is very predictable. There's no way to screw it up, and everyone is required to make a changeset as long as they're touching production code. Each version gets checked in. The change that was made cannot live alongside our current release process as it exists first because it will simply get replaced by the automation with real version numbers (and thus be pointless), second because we release off of main and don't have a bunch of branches that are storing the intermediate state of the repo at each release which means you can't just not track versions on main, and third because the consequences of not tracking versions would be that no one knows what's going on with the code that's being published on npm because it doesn't match the state of version control. So I could approve this but keep in mind it's just going to get replaced by our automation anyways, and doesn't work with our current release workflows.

.changeset/rare-garlics-kneel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@web5/api": minor
```
shamilovtim, line 2: shouldn't this be a patch? is it a breaking change? I believe with minor versions we have to get devrel involved

thehenrytsai: Taking a dependency on the latest version of `dwn-sdk-js` is a breaking change and we do want to get dev-rel involved.

shamilovtim (approved):
approving but the workspace:* stuff probably needs to get overwritten by the next release run

packages/api/package.json
thehenrytsai: @shamilovtim, I apologize for my stupidity, I read your response above multiple times and still don't get the full thing, I get the main point being: `workspace:*` does not work with our current release workflows with 3 reasons. Assuming we just need one of them to be true to prove `workspace:*` is a bad idea, let's just look at the first one:

> it will simply get replaced by the automation with real version numbers

I don't think this is true (but can be proven wrong easily of course by just getting this merged and see what happens), it'd be odd to me for the changeset-bot create a PR that replaces `workspace:*` to specific versions, because that kinda defeats the point this feature... but of course would validate your point.

I don't fully understand your point 2 and 3, but can't say I agree based on my interpretation.

thehenrytsai: No additional comments, so will merge and see what changeset-bot does.

csuwildcat (approved):

PR #695, opened 2024-06-22
Use empty filters for sync `EventsQuery`

LiranCohen:
Hope you don't mind I included the `ResumableTaskStore` in the DWN scaffolding to get the tests to pass, not sure if there is any agent-specific paths to test with it.

The fix you needed was to pass an empty filters array (to get all events) for sync. In future enhancements it will pass filters depending on what it's syncing and the permissions it has.

This is reliant on: https://github.com/TBD54566975/dwn-server/pull/137

Testing against that server version goes all green, so just need to merge that + bump the `dwn-server` dependency on this repo.


thehenrytsai (approved):
🙏 🐐

PR #713, opened 2024-06-29
wip: wallet connect

shamilovtim:


packages/agent/src/oidc-v2.ts
```diff
@@ -0,0 +1,690 @@
...
+    enc : 'XC20P',
+    typ : 'JWT',
+    kid : providerDidKid,
+  };
+
+  const additionalData = Convert.object({
+    ...protectedHeader,
+    // pin: randomPin({ length: 4 }),
+    // for testing
+    pin: '1234',
```
shamilovtim, line 653: need to revert this after finding a better testing strategy

shamilovtim: also need an e2e test where the pin is wrong

packages/agent/src/oidc-v2.ts
```diff
@@ -0,0 +1,689 @@
...
+ * Gets the correct OIDC endpoint out of the {@link OidcEndpoint} options provided.
+ * Handles a trailing slash on baseURL
+ *
+ * @param {Object} options the options object
+ * @param {string} options.baseURL for example `http://foo.com/connect/
+ * @param {OidcEndpoint} options.endpoint the OIDC endpoint desired
+ * @param {string} options.authParam this is the unique id which must be provided when getting the `authorize` endpoint
+ * @param {string} options.tokenParam this is the random state as b64url which must be provided with the `token` endpoint
+ */
+function buildOidcUrl({
```
thehenrytsai, line 206: Design pattern: not seeing the benefit of "data coupling" here given that half of the params are contextual.

Would prefer to have 4 independent methods with no coupling. Further still, most of the URL can just be constructed in-line within the caller, since no one is reusing the specific URL constructions.

packages/agent/src/oidc-v2.ts
```diff
@@ -0,0 +1,689 @@
...
+  const responseObject: HybridAuthResponse = {
+    ...options,
+    iat : currentTimeInSeconds,
+    exp : currentTimeInSeconds + 600, // Expires in 10 minutes.
+  };
+
+  return responseObject;
+}
+
+async function signJwt({
```
thehenrytsai, line 401: Sanity: can we expose/reuse existing lib in web5 packages? e.g. `Jwt.sign()` in `credentials` package?

packages/agent/src/oidc-v2.ts
```diff
@@ -0,0 +1,689 @@
...
+      data: Convert.string(`${header}.${payload}`).toUint8Array(),
+    });
+
+    const signatureBase64Url = Convert.uint8Array(signature).toBase64Url();
+
+    const jwt = `${header}.${payload}.${signatureBase64Url}`;
+
+    return jwt;
+  } catch (e) {
+    console.error(e);
```
thehenrytsai, line 429: Should not eat all errors.

packages/agent/src/connect-v2.ts
```diff
@@ -0,0 +1,199 @@
...
+}: WalletConnectOptions) {
+  // ephemeral client did for ECDH, signing, verification
+  // TODO: use separate keys for ECDH vs. sign/verify. could maybe use secp256k1.
+  const clientDid = await DidDht.create();
+
+  // Hash the code verifier to use as a code challenge and to encrypt the Request Object.
+  const { codeVerifieru8a } = Oidc.generateRandomCodeVerifier();
+
+  // Derive the code challenge based on the code verifier
+  const { codeChallengeu8a, codeChallengeb64url } =
```
thehenrytsai, line 27: Naming convention (`camelCasing`) and bad naming (`u8a`).

```suggestion
  const { codeChallengeBytes, codeChallengeBase64Url } =
```

packages/agent/src/utils.ts
```diff
@@ -84,4 +84,89 @@ export async function getPaginationCursor(message: RecordsWriteMessage, dateSort
...
+
+/**
+ * Polling function with interval, TTL accepting a custom fetch function
+ * @template T - the return you expect from the fetcher
+ * @param fetchFunction an http fetch function
+ * @param [interval=3000] how frequently to poll
+ * @param [ttl=300_000] how long until polling stops
+ * @returns T - the result of fetch
+ */
+export function pollWithTTL(
```
thehenrytsai, line 110: Convention:

`pollWithTtl`, yeah, it's kinda ugly in this case but convention wins, consider `pollUntilTimeout` instead. In fact, `dwn-sdk-js` has exported a very similar method called `pollUntilSuccessOrTimeout` if you are interested.

packages/agent/src/connect-v2.ts
```diff
@@ -0,0 +1,199 @@
...
+    });
+    if (!parResponse.ok) {
+      throw new Error(`${parResponse.status}: ${parResponse.statusText}`);
+    }
+
+    const parData: PushedAuthResponse = await parResponse.json();
+
+    // a deeplink to a web5 compatible wallet. if the wallet scans this link it should receive
+    // a route to its web5 connect provider flow and the params of where to fetch the auth request.
+    const walletURI = new URL('web5://connect/');
```
thehenrytsai, line 95: Nit: convention: `walletUri`

thehenrytsai (commented):
Reviewed until the line below:

`const parResponse = await fetch(pushedAuthorizationRequestEndpoint,`

where the code throws exception in test, so code beyond has no test coverage so unsure if it's review ready/worthy?

Will try to continue but figured I should publish comments so far to parallelize.

packages/agent/src/connect-v2.ts
thehenrytsai, line 1: Definitely warrants creating a `changeset`.

packages/agent/src/oidc-v2.ts
```diff
@@ -0,0 +1,689 @@
...
+export type PushedAuthResponse = {
+  request_uri: string;
+  expires_in: number;
+};
+
+/**
+ * An auth request that is compatible with both Web5 Connect and (hopefully, WIP) OIDC SIOPv2
+ * The contents of this are inserted into a JWT inside of the {@link PushedAuthRequest}.
+ */
+export type HybridAuthRequest = SIOPv2AuthRequest & Web5ConnectRequest;
```
thehenrytsai, line 46: Consider alternative: reads as "web5 connect request is a SIOPv2 auth request + some other custom things".

```suggestion
export type Web5ConnectRequest = SIOPv2AuthRequest & {
  /** PermissionGrants that are to be sent to the provider */
  permission_requests: ConnectPermissionRequests;
};
```

packages/agent/src/connect-v2.ts
```diff
@@ -0,0 +1,199 @@
...
+  /**
+   * The Web5 API provides a URI to the wallet based on the `walletUri` plus a query params payload valid for 5 minutes.
+   * The link can either be used as a deep link on the same device or a QR code for cross device or both.
+   * The query params are `{ request_uri: string; code_challenge: string; }`
+   * The wallet will use the `request_uri to contact the intermediary server's `authorize` endpoint
+   * and pull down the {@link HybridAuthRequest} and use the `code_challenge` to decrypt it.
+   *
+   * @param uri - The URI returned by the web5 connect API to be passed to a provider.
+   */
+  onUriReady: (uri: string) => void;
```
thehenrytsai, line 166: Can we rename to `onRequestUriReady` for extra clarity:
```suggestion
  onRequestUriReady: (uri: string) => void;
```

eslint.config.cjs
```diff
@@ -1,92 +1,91 @@
...
+    ignores: ["**/*.js", "**/*.cjs", "**/*.mjs"],
   },
-  plugins: {
-    '@typescript-eslint': tsPlugin,
-    'mocha': mochaPlugin
+  {
+    files: ["**/*.ts"],
+    rules: {
+      "no-undef": "off",
+      "no-redeclare": "off",
```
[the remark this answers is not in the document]
shamilovtim: The Ts compiler handles these errors correctly while eslint is dumber and throwing when it shouldn't. When I looked into the situation it turned out that you're not supposed to have these rules enabled for TS due to overlap with the compiler

packages/agent/src/oidc-v2.ts
```diff
@@ -0,0 +1,689 @@
...
+//   client_metadata?: { client_uri?: string; [key: string]: any };
+//   code_verifier?: string;
+//   nonce: string;
+//   redirect_uri: string;
+//   state?: string;
+//   [key: string]: any;
+// }
+
+/** An auth response that is compatible with both Web5 Connect and (hopefully, WIP) OIDC SIOPv2 */
+export type HybridAuthResponse = SIOPv2AuthResponse & Web5ConnectAuthResponse;
```
thehenrytsai, line 151: Consider similar changes as suggested for `HybridAuthRequest`

packages/agent/src/oidc-v2.ts
```diff
@@ -0,0 +1,689 @@
...
+// export interface AuthorizationRequestObject {
+//   claims?: { id_token: { delegation_grants: any; [key: string]: any } };
+//   client_id: string;
+//   client_metadata?: { client_uri?: string; [key: string]: any };
+//   code_verifier?: string;
+//   nonce: string;
+//   redirect_uri: string;
+//   state?: string;
+//   [key: string]: any;
+// }
```
thehenrytsai, line 148: Remove.

packages/agent/src/oidc-v2.ts
```diff
@@ -0,0 +1,689 @@
...
+  /** b64url encoded nonce used to associate a Client session with an ID Token, and to
+   * mitigate replay attacks. */
+  nonce?: string;
+  /** Custom claims. */
+  [key: string]: any;
+};
+
+/** The fields for an Web5 Connect Auth Response */
+export type Web5ConnectAuthResponse = {
+  delegation_grants: any[];
```
thehenrytsai, line 180: `delegatedGrants` to align with casing style of `privateKeyJwks`.

packages/agent/src/oidc-v2.ts
```diff
@@ -0,0 +1,689 @@
...
+  privateKeyJwks: Jwk[];
```
thehenrytsai, line 181: For my understanding: what's the reason of sending multiple private keys and how is it correlated to the grants created? It would be a bit odd if the agent has to manage an identity per grant.

packages/agent/src/oidc-v2.ts
```diff
@@ -0,0 +1,689 @@
...
+    const jwt = `${header}.${payload}.${signatureBase64Url}`;
+
+    return jwt;
+  } catch (e) {
+    console.error(e);
+  }
+}
+
+/** Take the decrypted JWT and verify it was signed by its public DID. Return parsed object. */
+async function verifyJwt({ jwt }: { jwt: string }) {
```
thehenrytsai, line 434: Similar comment to `signJwt()`.

packages/agent/src/oidc-v2.ts
```diff
@@ -0,0 +1,689 @@
...
+  }
+}
+
+/**
+ * Generates a random string value used to associate a Client authorization
+ * request with the Identity Provider's response callback.
+ *
+ * @returns A random state as a Base64Url encoded string.
+ */
+function generateRandomState() {
```
thehenrytsai, line 258: Consider generalizing for reuse and readability:

```suggestion
function generateRandomBase64Url(byteLength: number):  { ... }
```

packages/agent/src/oidc-v2.ts
```diff
@@ -0,0 +1,689 @@
...
+  const randomStateu8a = utils.randomBytes(12);
+  const ramdomStateb64url = Convert.uint8Array(randomStateu8a).toBase64Url();
+
+  return { randomStateu8a, ramdomStateb64url };
```
thehenrytsai, line 262: Typo `ramdom`. Also fix naming/casing:

```suggestion
  return { randomStateBytes, ramdomStateBase64Url };
```

packages/agent/src/connect-v2.ts
shamilovtim: It's not the request uri it's the wallet uri.

shamilovtim: In fact the request uri is in the payload of the wallet URI.

packages/agent/tests/connect-v2.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+  // });
+  // await testHarness.agent.identity.manage({
+  //   portableIdentity: await careerIdentity.export(),
+  // });
+  // const careerVm = await testHarness.agent.did.getSigningMethod({
+  //   didUri: careerIdentity.did.uri,
+  // });
+
+  describe('client connect integration tests', () => {
+    it('intiaties client wallet connect using a temporary clientDid', async () => {
```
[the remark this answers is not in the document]
shamilovtim: This test was just used  as a runner. I'm adding tests

packages/agent/src/connect-v2.ts
```diff
@@ -0,0 +1,199 @@
...
+    // a deeplink to a web5 compatible wallet. if the wallet scans this link it should receive
+    // a route to its web5 connect provider flow and the params of where to fetch the auth request.
+    const walletURI = new URL('web5://connect/');
+    walletURI.searchParams.set(
+      'nonce',
+      Convert.uint8Array(nonce).toBase64Url()
+    );
+    walletURI.searchParams.set('request_uri', parData.request_uri);
+    walletURI.searchParams.set('client_did', clientDid.uri);
+    walletURI.searchParams.set('code_challenge', codeChallengeb64url);
```
thehenrytsai, line 102: Sanity: I am under the impression that only `request_uri` and `client_id` are used to form the wallet URI. The rest are fetched directly from the Request URI itself?

packages/agent/src/connect-v2.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    // a deeplink to a web5 compatible wallet. if the wallet scans this link it should receive
+    // a route to its web5 connect provider flow and the params of where to fetch the auth request.
+    const walletURI = new URL('web5://connect/');
+    walletURI.searchParams.set(
+      'nonce',
+      Convert.uint8Array(nonce).toBase64Url()
+    );
+    walletURI.searchParams.set('request_uri', parData.request_uri);
+    walletURI.searchParams.set('client_did', clientDid.uri);
```
thehenrytsai, line 101: Sanity: did you mean `client_id`?

shamilovtim: no, I meant clientDid

packages/agent/src/connect-v2.ts
shamilovtim: not the case. the code challenge is necessary.

packages/agent/src/oidc-v2.ts
shamilovtim: there can be multiple private keys for a single DID. if that's the case then we'd be intentionally not transmitting existing private key material which would get lost forever.

packages/agent/tests/connect-v2.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+      portableIdentity: await alice.export(),
+    });
+  });
+
+  after(async () => {
+    await testHarness.clearStorage();
+    await testHarness.closeStorage();
+  });
+
+  // const careerIdentity = await testHarness.agent.identity.create({
```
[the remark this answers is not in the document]
shamilovtim: test isnt ready for review sorry. will get it fixed ASAP

packages/api/src/web5.ts
```diff
@@ -1,68 +1,27 @@
...
-}
+};
 
 /** Optional overrides that can be provided when calling {@link Web5.connect}. */
 export type Web5ConnectOptions = {
-
   /**
-   * When specified, wallet connect flow interacting with an external wallet would be triggered.
+   * When specified, external wallet connect flow is triggered.
+   * @remarks This param currently will not work in apps that were previously connected. It must be invoked at registration.
```
shamilovtim, line 24: comment to fix this comment to be a little more clear

packages/api/src/web5.ts
```diff
@@ -239,69 +205,72 @@ export class Web5 {
...
+
+        if (walletConnectOptions) {
+          WalletConnect.initClient({
+            connectServerUrl : walletConnectOptions.connectServerUrl,
+            walletUri        : walletConnectOptions.walletUri,
+            onUriReady       : walletConnectOptions.onUriReady,
+            permissionRequests:
+            walletConnectOptions.permissionRequests,
+            validatePin: walletConnectOptions.validatePin,
           });
```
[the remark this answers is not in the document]
shamilovtim: I liked how this was self documenting (even if pointless) for the reader of the code but at the same time your point is solid. Will see about it

packages/agent/src/oidc-v2.ts
```diff
@@ -0,0 +1,689 @@
...
+ * should be created for every authorization request.
+ *
+ * @see {@link https://datatracker.ietf.org/doc/html/rfc7636#section-4.1 | RFC 7636, Client Creates a Code Verifier}
+ *
+ * @returns A random code verifier as a Base64Url encoded string.
+ */
+function generateRandomCodeVerifier() {
+  const codeVerifieru8a = utils.randomBytes(32);
+
+  return { codeVerifieru8a };
```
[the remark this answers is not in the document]
shamilovtim: Normally I'd agree with what you're saying but because this is a specification it takes a lot more effort and research to find a spec based number like 32 than it does to just use a bespoke method that enshrines the spec into the code so I'd rather keep this. No problem on the name change

packages/agent/src/oidc-v2.ts
```diff
@@ -0,0 +1,689 @@
...
+}
+
+/**
+ * Generates a cryptographically random key called a "code verifier" in
+ * accordance with the RFC 7636 PKCE specification. A unique code verifier
+ * should be created for every authorization request.
+ *
+ * @see {@link https://datatracker.ietf.org/doc/html/rfc7636#section-4.1 | RFC 7636, Client Creates a Code Verifier}
+ *
+ * @returns A random code verifier as a Base64Url encoded string.
```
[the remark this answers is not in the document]
shamilovtim: Thanks for catching will fix comment

packages/agent/src/connect-v2.ts
```diff
@@ -0,0 +1,199 @@
...
+      const jwt = await Oidc.decryptAuthResponse(clientDid, jwe, pin);
+      const verifiedAuthResponse = (await Oidc.verifyJwt({
+        jwt,
+      })) as HybridAuthResponse;
+
+      // return the grants for liran to
+      console.log(verifiedAuthResponse);
+    }
+  } catch (e) {
+    console.error(e);
```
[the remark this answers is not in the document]
shamilovtim: I don't agree that this is swallowing the error; it logs it. I could return the error if that's what you mean?

packages/agent/src/connect-v2.ts
```diff
@@ -0,0 +1,199 @@
...
+  // ephemeral client did for ECDH, signing, verification
+  // TODO: use separate keys for ECDH vs. sign/verify. could maybe use secp256k1.
+  const clientDid = await DidDht.create();
+
+  // Hash the code verifier to use as a code challenge and to encrypt the Request Object.
+  const { codeVerifieru8a } = Oidc.generateRandomCodeVerifier();
+
+  // Derive the code challenge based on the code verifier
+  const { codeChallengeu8a, codeChallengeb64url } =
+    await Oidc.deriveCodeChallenge(codeVerifieru8a);
```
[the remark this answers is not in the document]
shamilovtim: It's self documenting. It makes it clear that the code challenge is derived from the code verifier. The other way one method would spit out 3 magical values and a user would have to dig to understand what's happening.

packages/agent/src/oidc-v2.ts
shamilovtim: As someone who got to use this method in multiple places it was so helpful to me to centralize the routes and understand the order they're used it and the params that belong to each route. Inlining would never do that. Splitting into 4 is just a mess. The coupling is the point, it's helpful for understanding.

packages/agent/src/oidc-v2.ts
shamilovtim: Point taken I can look into replacing if it's a 1:1 replacement but it's worth pointing out that everything here works for Daniel's demo right now as is

packages/agent/src/oidc-v2.ts
shamilovtim: I can return an error

packages/agent/src/connect-v2.ts
shamilovtim: agree on bytes but isn't the convention for base64url `b64url`?

packages/agent/src/connect-v2.ts
shamilovtim: well yeah but I haven't made on yet :)

packages/agent/src/authServer.js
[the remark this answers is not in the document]
shamilovtim: dwn-server is a separate repo in the node modules folder with es6 classes and abstractions. this is a directly editable, readable test script. i can move it out to a different folder, it probably shouldn't be where it's at now

packages/api/src/web5.ts
```diff
@@ -239,69 +205,72 @@ export class Web5 {
...
-
-          // The User Agent will manage the Identity, which ensures it will be available on future
-          // sessions.
-          await userAgent.identity.manage({ portableIdentity: await identity.export() });
-
-        } else if (existingIdentityCount === 1) {
-          // An existing identity was found in the User Agent's tenant.
-          identity = identities[0];
-
         } else {
```
shamilovtim, line 260: note for review in case it's not clear the impersonation isn't here yet. i need to work with liran to add it

eslint.config.cjs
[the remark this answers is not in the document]
thehenrytsai: I see... though it appears this updated eslint config actually causes more error (see CICD log failing immediately in `common` pacakge), I tried with the original config, no errors. You'll need to investigate, please leave inline comment if this turns out to be necessary so that this valuable tribal context knowledge is not lost.

packages/agent/src/connect-v2.ts
thehenrytsai: Did a string search in the repo, "base64url" occurred 400+ times, while b64url only showed up in this PR, so "Base64Url" is the convention.

packages/api/src/web5.ts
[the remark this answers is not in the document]
thehenrytsai: 🙏

packages/agent/src/connect-v2.ts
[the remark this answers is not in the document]
thehenrytsai: Hiding implementation detail whenever possible promotes readable code, a.k.a. Encapsulation.

In this case, what readers really need to know is the challenge, so a `generateCodeChallenge()` would suffice and more readable IMO. Not asking to have magical values, but if the details (code verifier what not) are "encapsulated" behind a method, it helps reader who doesn't care about the detail to ignore the detail, while allowing people who want to know about the detail drill in.

We are probably being way too nerdy about this! 🤣 

If you don't feel strongly about this either way, please change it. If you do, I am not willing to die on this hill. Full disclosure though, as a DRI, I will change it later if it ever bothers me enough.

packages/agent/src/connect-v2.ts
thehenrytsai: Sanity x2: not asking about `clientDid`, did you mean `client_did`? I don't see references to `client_did` in the context of SIOP with a quick search, but saw a tons of hits for `client_id`.

Are you able to send me a link on where `client_did` is described in spec? For my education. 🙏

packages/agent/src/connect-v2.ts
thehenrytsai: Sanity 2x plus eduction for me: can you explain and/or point me to documentation that says challenge and nonce much be encoded in the URI when they are already part of the Auth Request object? 🙏 

Seems silly to me that the point of Auth Request object is to alleviate the need to pass properties via query params, but you have to do it anyways.

packages/agent/src/connect-v2.ts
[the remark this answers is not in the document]
thehenrytsai: My apologies, you are right, this is not "technically" swallowing all errors, but "effectively" doing so, because the caller would not be able to distinguish between success and failure.

This is typically a definite no-no. Happy to discuss further If you think this warrants exception, else please throw/don't bother catch.

packages/agent/src/connect-v2.ts
thehenrytsai: Thanks for correcting me, which happens to be case in point LOL: dumb code readers like me get easily confused. So please qualify the URI. Perhaps `onWalletUriRead`, hopefully I get it more right this time.

packages/agent/src/oidc-v2.ts
thehenrytsai: Pretty sure you've got it right, just still need more explanation for my learning: can there be more than 1 key that controls a delegated identity (such as DID DHT)? Even if so, would there ever be a case where the wallet need to generate more than one private key per DID? Seems 
to be self-inflicted complexity.

Perhaps the array is more to do with generating multiple identities? If so perhaps more "structure" to couple the grants with identity/DID/private key would be useful?

@csuwildcat or @LiranCohen can you comment on actual usecase?

packages/agent/src/connect-v2.ts
shamilovtim: `client_did` is *our* param. it's part of web5 connect. the `client_id` that you're referring to is a website. it's described in the type definitions

packages/agent/src/oidc-v2.ts
thehenrytsai: > understand the order they're used it and the params that belong to each route

Got it, I see your intent now why you documented 1. ~ 4. steps here inline. The documentation is awesome and super useful for sure, but it feels misplaced for two reasons:
1. Describing input output of each URL for a function called `buildOidcUrl` feels out of place, reader normally wouldn't be expecting to see param info decribed here.
1. This info would probably be more discoverable and useful at the top of the entry point of (`connect-v2.ts`).

 I disagree quite strongly on "Splitting into 4 is just a mess" statement.

It is undeniably useful for the developer (you) of the feature to structure the code this way during development. But for the reader of this function (me), it is not obvious what params can be used/optional/required/in-combination with which endpoint without reading the documentation or code carefully (not self-documenting, quoting from your response in another comment), coupled with the observation that this is not a utility function, lead me to think that the alternative is more desirable than the coincidental coupling in code.

In my view this is an example of "coincidental coupling", which often makes the code look neat on the surface, but require avoidable context to understand.

Apologies for geeking out on this one, honestly there is not a lot of code here to be that confused! But hopefully I explained my principle in thinking. Bottom line is, if you don't feel strongly, please change it, if you feel strongly don't change it, I won't die on this hill either and block the PR on this. I can always change it later if it bothers me enough. 🙏

packages/agent/src/connect-v2.ts
shamilovtim: > Seems silly to me that the point of Auth Request object is to alleviate the need to pass properties via query params, but you have to do it anyways.

It's a choice that was made to encrypt the auth response with the code challenge because we don't do server side validation like a real IDP would, because we don't have a real server and we aren't a real IDP. In the pure OAuth flow there'd be a way of authenticating the person fetching from the `request_uri`.

shamilovtim: Keep in mind that in the future there might be transmission mechanisms that don't require an intermediary server at all. We want to future proof for that and not go all in on server side validation. Right now our server is "dumb" and we can keep it that way with this approach.

packages/agent/src/oidc-v2.ts
shamilovtim: done

packages/agent/src/oidc-v2.ts
shamilovtim: No need for this, it's the way it is because its locked to 12 bytes

packages/agent/src/oidc-v2.ts
shamilovtim: > can there be more than 1 key that controls a delegated identity (such as DID DHT)

Yeah a DID can have signing, verification and key agreement keys, and so on. In fact Frank is the one who encouraged me to have multiple keys for a DID and you can see it in the DIDs he frequently creates around the SDK he uses ed25519 for signing and secp256k1 for key agreement (I believe).

I don't disagree that it's complex. But I think I'd rather have a comprehensive complexity burn down rather than start to hammer away at complexity as part of a review of a typedef in a PR, otherwise I'm going to be burning down complexity rather than getting this merged. If timeline permits I'm happy to chip away at this though.

shamilovtim: > Perhaps the array is more to do with generating multiple identities? If so perhaps more "structure" to couple the grants with identity/DID/private key would be useful?

Unrelated. Each one of those can have multiple private keys too. So not conflating those two yet

packages/agent/src/oidc-v2.ts
shamilovtim: done

packages/agent/src/oidc-v2.ts
shamilovtim: done

packages/agent/src/oidc-v2.ts
shamilovtim: done

packages/agent/src/connect-v2.ts
shamilovtim: well there's a slight hitch I already ask the user for something called `walletUri` 

`onUriReady` is where I take the `walletUri` they gave me and add the query params to it. any thoughts on a better approach?

packages/agent/src/oidc-v2.ts
shamilovtim: done

packages/agent/src/connect-v2.ts
shamilovtim: done

packages/agent/src/utils.ts
shamilovtim: done

packages/agent/src/connect-v2.ts
shamilovtim: done

packages/agent/src/oidc-v2.ts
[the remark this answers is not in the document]
shamilovtim: done

packages/agent/src/connect-v2.ts
```diff
@@ -0,0 +1,199 @@
...
+export type WalletConnectOptions = {
+  /** The URL of the intermediary server which relays messages between the client and provider */
+  connectServerUrl: string;
+
+  /**
+   * The URI of the Provider (wallet) which is used to generate the URI returned in `onUriReady`
+   * e.g. `web5://` or `http://localhost:3000/`.
+   *
+   */
+  walletUri: string;
```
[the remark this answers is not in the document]
shamilovtim: lets move conversation here and resolve this one: https://github.com/TBD54566975/web5-js/pull/713#discussion_r1678626819

eslint.config.cjs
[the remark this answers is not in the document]
shamilovtim: will do. i'll look into the extra errors but i doubt they're from turning rules **off** they're probably from something else i did

packages/agent/src/connect-v2.ts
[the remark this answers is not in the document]
shamilovtim: So as the reviewer of this you got to see that the derivation was from a param passed into the function. You don't need to click into the functions to understand what's happening. In your suggested approach I do everything opaquely in one function and then spit out three values, and then the reviewer or reader needs to climb through multiple levels of nested functions in order to understand what's going on.

>  Full disclosure though, as a DRI, I will change it later if it ever bothers me enough.

Of course this is trivial code but I don't think this is how a disagreement should be handled. If you're the sole maintainer knock yourself out at that point it's your thing.

packages/agent/src/oidc-v2.ts
shamilovtim: Ok but let me call out: this is the only place outside of the `dwn-server` repo where the routes are centralized, constructed, and then returned.  A comment that does the same thing would be double the code. It would be 4 methods plus a giant comment the size of them. When instead we have a method that accomplishes both tasks. The method is simple to understand and requires no special context. We could add 4 overloads but I had decided it wasn't worth it...

shamilovtim: > honestly there is not a lot of code here to be that confused

Right that's exactly what I'm saying. It's easily readable to the extent that there's not going to be any confusion as to what it does. No one who knows how to write code is going to be confused by this method, so it shouldn't be a target for some sort of design pattern optimization.

packages/agent/src/connect-v2.ts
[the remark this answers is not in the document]
shamilovtim: Oh so you're saying you want me to throw. Wasn't clear from what you were asking. I'll go through and clean this up

packages/agent/src/oidc-v2.ts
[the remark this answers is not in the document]
shamilovtim: fixed

packages/agent/src/connect-v2.ts
thehenrytsai: Sorry, sanity x3: I understand custom claims are fine, but isn't SIOPv2 using `client_id` for specifying the client DID?

e.g. https://openid.net/specs/openid-connect-self-issued-v2-1_0.html#section-7.2.3

So why invent another field for it?

packages/agent/src/oidc-v2.ts
thehenrytsai: My apology for being a little confused when reading how the code works, not much though, just little, hence the suggestion. But seems like you feel pretty strongly about this one, so we'll stop at this.

packages/agent/src/connect-v2.ts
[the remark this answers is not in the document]
thehenrytsai: In my suggested approach, I'd spit out 2 values not 3, which is my point: as a reviewer, only 2 values are interesting here, hence the encapsulation suggestion of hiding other details.

Sounds like you feel strongly about this one, which is totally okay, I concede and resolve.

Though to be clear, when there is a deadlocked disagreement, DRI is precisely the way to tiebreak, not the case here as I am not willing to die here today. If I do make further refactoring in the future, it'll undergo the same review process and consideration by others like any PR, not a big deal, my bad for calling out the disclosure, I was just hoping that you won't think of me as an a-----e if I ever do attempt to refactor.

packages/agent/src/connect-v2.ts
shamilovtim: It's a field that accommodates either a DID or a callback URL. We have a callback URL. we already use 'client_id', it's just not in the query params. That being said I think we can get rid of the client did param altogether, as I already get it from the JWK

packages/agent/src/connect-v2.ts
shamilovtim: Done

packages/agent/src/connect-v2.ts
shamilovtim: Going to do more research on this, lets leave it open

packages/agent/src/connect-v2.ts
shamilovtim: update:
- removed the nonce from the query params
- going to remove `client_did` from the query params

code challenge will remain but can be tabled for a discussion about the future web5 connect spec and whether encrypting with the `code_challenge` was truly the right decision, and we can evolve the spec to be what we want with input from @frankhinek @csuwildcat .

packages/agent/src/authServer.js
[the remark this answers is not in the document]
thehenrytsai: Being able to edit this file is not necessarily a good thing, the flow works against this script gives no guarantee that it will work against the actual DWN server. Making sure two copies behaving exactly same in this case is likely very impractical and thus asking for trouble.

Regarding testability, if you want to do unit/functional testing, then mocking auth server calls should be sufficient. If you want more scenario/e2e tests that rely on real server interactions, I suggest to invoke the "real" auth server code: I believe there are already tests in this repo that directly depend on a running `dwn-server` instance, why not piggyback on the same pattern.

If you really want to have white-box control of the auth server here, perhaps generalize auth server code as a library in one of the `web5` packages and have the `dwn-server` import that package. But I am not convinced this is a good idea because you'll still end up having to duplicate the entire routing code (express).

packages/agent/src/oidc-v2.ts
thehenrytsai: For my learning:

1. Is state required to be random bytes? This is different to what I normally understood to be the intent of `state` in OAuth/OIDC.
1. Is the 12 bytes a spec compliance thing, or something that has enough entropy?
1. Is the `nonce` required to be derived `state`?

packages/agent/src/authServer.js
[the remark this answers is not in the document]
shamilovtim: its just for local development, not for automated testing. but sure i can remove it.

packages/agent/src/connect-v2.ts
shamilovtim: should be good to close this out and make a ticket about encrypting with the `code_challenge`

packages/agent/src/connect-v2.ts
shamilovtim: Ok you were right. We already got the callback endpoint from the redirect URI and we don't need to use `client_id` for it. Sorry, and nice catch!

shamilovtim: fixed

packages/agent/src/oidc-v2.ts
shamilovtim: Good catch actually. I misread the spec and it was poorly written:

> The nonce parameter value needs to include per-session state and be unguessable to attackers. One method to achieve this for Web Server Clients is to store a cryptographically random value as an HttpOnly session cookie and use a cryptographic hash of the value as the nonce parameter. In that case, the nonce in the returned ID Token is compared to the hash of the session cookie to detect ID Token replay by third parties. A related method applicable to JavaScript Clients and other Browser-Based Clients is to store the cryptographically random value in HTML5 local storage and use a cryptographic hash of this value.

By "include per-session state" they just mean the nonce should be a nonce; which is tautological and unnecessary to write. I solved for your feedback and removed all of the abstraction here.

shamilovtim (commented):
fixed up another set of change requests

packages/agent/src/oidc-v2.ts
thehenrytsai: Sounds good, please create an issue and reference it in a TODO comment here so we don't loose track.

packages/agent/src/oidc-v2.ts
thehenrytsai: So for my understanding: is the 12 bytes a spec compliance thing?

If it's not a spec thing, then seems odd to choose 12. Conventionally 16 or 32 bytes (128 or 256 bits) are used.

packages/agent/src/oidc-v2.ts
```diff
@@ -307,17 +260,18 @@ async function createAuthRequest(
   >
 ) {
   // Generate a random state value to associate the authorization request with the response.
-  const { randomStateBytes, randomStateBase64Url } = generateRandomState();
+  const stateBytes = utils.randomBytes(12);
 
   // Generate a random nonce value to associate the ID Token with the authorization request.
-  const nonce = await deriveNonceFromInput(randomStateBytes);
+  const nonceBytes = await Sha256.digest({ data: utils.randomBytes(12) });
```
thehenrytsai, line 266: consider just:

```suggestion
  const nonceBytes = utils.randomBytes(32);
```

The `randomBytes()` is called inline, I don't see functional benefit or difference wrapping it with a `Sha256.digest()` call, it add more unanswered questions.

packages/agent/src/oidc-v2.ts
```diff
@@ -297,7 +250,7 @@ async function deriveCodeChallenge(codeVerifier: Uint8Array) {
 // TODO: when implementing pure OIDC split up the Web5 and OIDC params
```
thehenrytsai, line 250: Sanity: Is this still applicable? Obsolete? Remove?

packages/agent/src/oidc-v2.ts
```diff
@@ -196,73 +196,26 @@ function buildOidcUrl({
...
     /**  4. client gets {@link Web5ConnectAuthResponse */
     case 'token':
       return tokenParam
-        ? appendPathToUrl({
-          path : `token/${tokenParam}.jwt`,
-          url  : baseURL,
-        })
+        ? concatenateUrl(baseURL, `token/${tokenParam}.jwt`)
         : '';
     // TODO: metadata endpoints?
```
thehenrytsai, line 213: Suggesting throw over returning '';

packages/agent/src/oidc-v2.ts
shamilovtim: We hash because it's part of the specification. This was there before, has nothing to do with inline randombytes? Maybe I'm not understanding the review

packages/agent/src/oidc-v2.ts
shamilovtim: Still applies

packages/agent/src/connect-v2.ts
thehenrytsai: Thanks for removing `nonce` from query params.

I see that you also removed `client_id` (your `client_did`) from the URL, though, from examples that I can find in the specs on the internet, I always see `client_id` included in the URL showing that it is an exception to the rule when using `request_uri`. I am not going to insist to include `client_id` in the URL because you are right in saying that it is redundant, but it does "appear unconventional" to not include it. The rationale for having it in the URL seems so that it can act as a request lookup key/short-cut, but that's not how we implement `request_uri` generation.

You are good to resolve this comment thread if you prefer no further action on `client_id`. I'll create a separate thread on the topic of `code_challenge` to not bloat this thread any further!

packages/agent/src/connect-v2.ts
thehenrytsai: Final FYI:

I see that you removed `client_id` from the URL, though, from examples that I can find in the specs on the internet, I always see `client_id` included in the URL showing that it is an exception to the rule when using `request_uri`. I am not going to insist to include `client_id` in the URL because you are right in saying that it is redundant, but it does "appear unconventional" to not include it. The rationale for having it in the URL seems so that it can act as a request lookup key/short-cut, but that's not how we implement `request_uri` generation.

You are good to resolve this comment thread if you see no further action on `client_id`.

packages/agent/src/connect-v2.ts
```diff
@@ -0,0 +1,195 @@
...
+  // ephemeral client did for ECDH, signing, verification
+  // TODO: use separate keys for ECDH vs. sign/verify. could maybe use secp256k1.
+  const clientDid = await DidDht.create();
+
+  // Hash the code verifier to use as a code challenge and to encrypt the Request Object.
+  const { codeVerifierBytes } = Oidc.generateRandomCodeVerifier();
+
+  // Derive the code challenge based on the code verifier
+  const { codeChallengeBytes, codeChallengeBase64Url } =
+      await Oidc.deriveCodeChallenge(codeVerifierBytes);
```
thehenrytsai, line 28: The concern I have here is not necessarily with us encrypting the request object using a symmetric key then passing the key to the wallet so it can decrypt it (though I am curious if this pattern is something we invented in web5, or a known OIDC pattern. The comment here from code review stand point is that the code makes it unnecessarily confusing for the reader to conclude that's what the code is doing. 

The code seems to conflate with the intent of `code_challenge` and `code_verifier`. The `code_verifier` is not persisted anywhere for later verification, so in the end all the fancy names and extra code just serve the purpose of creating a 32 byte random number no?

If that's the case, why don't we just do `const symmetricKey = utils.randomBytes(32);` and call it a day?

packages/agent/src/oidc-v2.ts
thehenrytsai: Sorry I don't understand the TODO, can you elaborate/clarify? I thought what you are doing is ODIC compliant already? If so, what does it mean by "when implementing pure OIDC"?

packages/agent/src/oidc-v2.ts
thehenrytsai: I am not convinced this is REQUIRED by the specification, can you point me to where in the spec that says this? 

> This was there before

I assume you mean this was part of the code you were given, regardless, it still appears unnecessary to me, you are not reusing the random bytes anywhere, why start with a value with lower entropy when you can just go straight to the higher entropy random value and call it a day.

packages/agent/src/oidc-v2.ts
shamilovtim: We're doing web5 connect in this PR. The comment says that when doing a PR for OIDC to add more separation of concerns between web5 connect and OIDC.

shamilovtim: There are other PRs that will come in the future that will refine our code for use with OIDC4VC SIOPv2 rather than only Web5 Connect. The web5 connect PR is tested on web5 connect only, it is not tested with third party OIDC4VC SIOPv2 flows. There will be more refinement to work with both protocols.

packages/agent/src/connect-v2.ts
shamilovtim: > I see that you removed client_id from the URL, though, from examples that I can find in the specs on the internet, I always see client_id included in the URL showing that it is an exception to the rule when using request_uri

We are using a PAR so we don't need to include it in query params, it's in the message body

eslint.config.cjs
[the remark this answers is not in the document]
shamilovtim: I fixed the errors here. Keep in mind the eslint recommended config is not for Typescript but for javascript:
Screenshot 2024-08-05 at 4 14 23 PM

That being said I don't want to spend much more time on this since I expect we'll probably move to Biome or a better option than bloated eslint

examples/dignal.html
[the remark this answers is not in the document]
shamilovtim: It has frontend features in it. I'm just keeping the UI in place and I'll clean it up in a later PR

packages/agent/src/authServer.js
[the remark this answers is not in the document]
shamilovtim: done

packages/agent/tests/connect-v2.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+      const foo = await WalletConnect.initClient({
+        walletUri          : 'http://localhost:8080/',
+        connectServerUrl   : 'http://localhost:8080/connect',
+        permissionRequests : {'12345': { protocolDefinition: {} as any, permissionScopes: {} as any }},
+        onUriReady         : (uri) => console.log('onUriReady callback called: ', uri),
+        validatePin        : async () => '1234'
+      });
+    });
+
+    // it("signs the OIDC request using the private key of the clientDid", async () => {});
```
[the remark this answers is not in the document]
shamilovtim: stale

packages/agent/tests/connect-v2.spec.ts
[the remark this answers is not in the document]
shamilovtim: stale

packages/agent/tests/connect-v2.spec.ts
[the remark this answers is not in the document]
shamilovtim: stale

packages/agent/src/oidc-v2.ts
shamilovtim: issue made

packages/agent/src/connect-v2.ts
shamilovtim: changeset in PR

packages/agent/src/oidc-v2.ts
shamilovtim: The DID specification allows for multiple private keys for a DID so I think this change request isn't necessary

packages/agent/src/oidc-v2.ts
shamilovtim: fixed

packages/agent/src/oidc-v2.ts
shamilovtim: fixed

packages/agent/src/oidc-v2.ts
shamilovtim: fixed

packages/agent/src/connect-v2.ts
shamilovtim: fixed

shamilovtim:
I fixed the codecov bot. My PR was increasing average coverage by `0.13%` but codecov was set to nitpick every change of every PR for every monorepo which means simple and harmless changes or WIP multi step PRs can cause a failing error status

codecov.yml
```diff
@@ -59,3 +59,4 @@ coverage:
     patch:
       default:
         informational: true # Don't gate PRs based on Codecov passing thresholds
+        if_ci_failed: success
```
shamilovtim, line 62: this `# Don't gate PRs based on Codecov passing thresholds` wasn't actually working until the line i added

packages/proxy-agent/package.json
```diff
@@ -71,7 +71,7 @@
   "dependencies": {
     "@web5/agent": "workspace:*",
     "@web5/common": "1.0.0",
-    "@web5/crypto": "1.0.0",
+    "@web5/crypto": "workspace:*",
```
shamilovtim, line 74: this was the only way i could get crypto to behave correctly on CI

eslint.config.cjs
```diff
@@ -1,92 +1,86 @@
-const eslint = require('@eslint/js');
-const globals = require('globals');
-const tsParser = require('@typescript-eslint/parser');
-const tsPlugin = require('@typescript-eslint/eslint-plugin');
-const mochaPlugin = require('eslint-plugin-mocha');
+const eslint = require("@eslint/js");
+const globals = require("globals");
+const tsParser = require("@typescript-eslint/parser");
+const tsPlugin = require("@typescript-eslint/eslint-plugin");
+const mochaPlugin = require("eslint-plugin-mocha");
```
shamilovtim, line 5: prettier did this

packages/agent/src/oidc.ts
```diff
@@ -0,0 +1,709 @@
...
+    /**  4. client gets {@link Web5ConnectAuthResponse */
+    case 'token':
+      if (!tokenParam)
+        throw new Error(
+          `tokenParam must be providied when building a token URL`
+        );
+
+      return tokenParam
+        ? concatenateUrl(baseURL, `token/${tokenParam}.jwt`)
+        : '';
```
thehenrytsai, line 223: Sanity: since there is a check `if(!tokenParam)` above, there is no need for this conditional return?

Same goes to `authParam` check above.

package.json
```diff
@@ -44,5 +45,6 @@
       "fast-xml-parser@<4.4.1": ">=4.4.1",
       "@75lb/deep-merge@<1.1.2": ">=1.1.2"
     }
-  }
+  },
+  "packageManager": "pnpm@9.6.0+sha256.dae0f7e822c56b20979bb5965e3b73b8bdabb6b8b8ef121da6d857508599ca35"
```
thehenrytsai, line 49: For my learning: why is this needed? Isn't `hermit` (not that I love it) what the team uses for the same purpose assuming I guessed the intent correctly?

package.json
```diff
@@ -17,9 +17,9 @@
   "scripts": {
     "clean": "pnpm npkill -d $(pwd)/packages -t dist && pnpm npkill -d $(pwd) -t node_modules",
     "build": "pnpm --recursive --stream build",
-    "dwn-server": "DWN_SERVER_PACKAGE_JSON=node_modules/@web5/dwn-server/package.json node node_modules/@web5/dwn-server/dist/esm/src/main.js || true",
```
thehenrytsai, line 20: For my understanding, why is this removed?

packages/agent/src/oidc-v2.ts
thehenrytsai: I see you've change to `PortableDid`, so closing.

package.json
shamilovtim: don't need it ever since the dwn-server bin got exposed by a PR i did

package.json
shamilovtim: this tells contributors what packageManager we're using. its an industry standard around corepack which is what hermit is using

packages/agent/src/connect-v2.ts
thehenrytsai: I see that you've move the verifier code into `generateCodeChallenge()`, which helps encapsulate the logic for better readability I guess, but my main question remains: the fact that the `code_verifier` is not persisted anywhere tells me something is half baked or doesn't add up here.

I'd suggest that we just do `const symmetricKey = utils.randomBytes(32);` and close out this comment and worry about implementing `code_challenge` and `code_verifier` properly in a subsequent PR.

But not okay for having a bunch of fancy properties/params for appears to add extra security or feature when it in the end is just used as a symmetric key.

Let's jump on a call if I am still missing the point and close out this comment.

examples/dignal.html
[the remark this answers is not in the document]
thehenrytsai: It looks like you've removed this. So I am going to revolve this now!

.github/workflows/docs-ci.yml
```diff
@@ -36,8 +36,8 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           report_changed_scope_only: false
-          fail_on_warnings: true
-          fail_on_error: true
```
thehenrytsai, line 40: These were intentionally turn on to keep the documentation in good shape.
We've fixed all past issues for a while now, let's keep it that way so we don't fall behind again and also have annoying warnings/errors polluting every single PR.

.github/workflows/tests-ci.yml
```diff
@@ -106,12 +106,6 @@ jobs:
         with:
           cache: "true"
 
-      - name: Print Node.js, npm, & pnpm versions for debugging if needed
-        run: |
-          node -v
-          npm -v
-          pnpm -v
-
```
thehenrytsai, line 114: For my learning: this was added recently to debug something (I don't recall what), judging from the removal, you probably concluded this is unnecessary, can you elaborate a bit?

package.json
thehenrytsai: That's awesome!

package.json
thehenrytsai: You had me at "industry standard", does this mean we need to remember to manually update this line every time we run hermit update pnpm? If so, you'd loose me again.

packages/agent/.mocharc.json
```diff
@@ -1,5 +1,6 @@
 {
   "enable-source-maps": true,
   "exit": true,
-  "spec": ["tests/compiled/**/*.spec.js"]
-}
\ No newline at end of file
+  "spec": ["tests/compiled/**/*.spec.js"],
+  "timeout": 10000
```
thehenrytsai, line 5: We shouldn't be doing this in principle, will read on, but in general we should do this at per test level if we have absolutely have no choice.

packages/agent/src/oidc.ts
```diff
@@ -0,0 +1,703 @@
...
+    contacts?: string[];
+    /** URI that points to a terms of service document for the client */
+    tos_uri?: string;
+    /** URI that points to a privacy policy document */
+    policy_uri?: string;
+  };
+};
+
+/**
+ * An auth request that is compatible with both Web5 Connect and (hopefully, WIP) OIDC SIOPv2
```
thehenrytsai, line 125: Sanity: remove `(hopefully, WIP)`? Same for `Web5ConnectAuthResponse`.

packages/agent/src/oidc.ts
```diff
@@ -600,18 +594,116 @@ function encryptAuthResponse({
...
+    const message = await dwn.processRequest({
+      author        : did,
+      target        : did,
+      messageType   : DwnInterface.RecordsWrite,
+      messageParams : {
+        recipient    : ephemeralDid.uri,
+        protocolPath : 'grant',
+        protocol     : ' https://tbd.website/dwn/permissions',
+        dataFormat   : 'application/json',
+        data         : Convert.object(permissionRequestData).toUint8Array(),
```
thehenrytsai, line 617: At a glance the data is missing `delegated: true`.

packages/agent/src/oidc.ts
```diff
@@ -600,18 +594,116 @@ function encryptAuthResponse({
...
+      description : 'The app is asking to Records Write to http://profile-protocol.xyz',
+      scope       : {
+        interface : DwnInterfaceName.Records,
+        method    : DwnMethodName.Write,
+        protocol  : 'http://profile-protocol.xyz',
+      },
+    };
+
+    // TODO: Confirm this
+    const message = await dwn.processRequest({
```
thehenrytsai, line 608: @LiranCohen would know how it's done at `web5` layer, but a protocol scoped grant needs to have a "protocol tag" associated with the record also.

packages/agent/src/oidc.ts
```diff
@@ -600,18 +594,116 @@ function encryptAuthResponse({
   return compactJwe;
 }
 
+export async function createPermissionGrants(
+  selectedDids: string[],
+  ephemeralDid: BearerDid,
```
thehenrytsai, line 593: Sanity: did you mean app DID or delegated DID?

packages/agent/src/oidc.ts
```diff
@@ -0,0 +1,703 @@
...
+  const jwt = Convert.uint8Array(decryptedJwtBytes).toString();
+
+  return jwt;
+}
+
+async function deriveSharedKey(
+  privateKeyDid: BearerDid,
+  publicKeyDid: DidDocument
+) {
+  const exportedClientDid = await privateKeyDid.export();
```
thehenrytsai, line 508: `exportedClientDid` is misleading in this generalized function as it is used by both sender and receiver to generate shared key in reverse manner. ie. in the case of `submitAuthResponse()`, the `privateKeyDid` represents the delegated DID, not the ephemeral client DID.

Consider rename to something generic like `privatePortableDid`.

package.json
shamilovtim: nope i didn't manually put this there it's put there by default by corepack

.github/workflows/tests-ci.yml
shamilovtim: this particular ci workflow is time sensitive. i tuned it to be as fast as possible in the past. so adding a bunch of random stuff that's on net not necessary hinders keeping it fast. plus worth pointing out all of these versions are well known throughout multiple places in the repo (like for example the hermit folder)

.github/workflows/docs-ci.yml
shamilovtim: the check is not working correctly for ESM imports. it also prevents us from moving fast. if it's polluting all of our PRs, lets shut the check off.

shamilovtim: what value is there in this being done because you're trying to satisfy a CI check so you can push code?

```
/** gets the permission grants */
getPermissionGrants()
```

i have seen this all over the repo and in some cases comments written by you because you're satisfying the check. i don't want to do this, it's double the code.

package.json
shamilovtim: pnpm would be upgraded by running `corepack up`. if we're somehow sidestepping that right now then yes this line would not be kept updated. but then i don't understand why we have corepack setup in the repo

packages/agent/src/oidc.ts
shamilovtim: this hasn't been validated with pure OIDC we had this same conversation about a different comment that you were asking me to remove. when we do both connect and OIDC (w/o connect) in the wallet I will remove this

packages/agent/src/oidc.ts
shamilovtim: good catch. changed

packages/agent/src/oidc.ts
shamilovtim: another good catch. changed

packages/agent/.mocharc.json
shamilovtim: ok removed

packages/agent/src/oidc.ts
shamilovtim: donw

packages/agent/src/oidc.ts
shamilovtim: lets mark this resolved and that can come in integration later

packages/agent/src/oidc.ts
shamilovtim: as with the last comment useful info but i think it needs to be part of a later PR rather than this one

packages/agent/src/connect-v2.ts
shamilovtim: It's a good call out but the suggestion I think misses the mark. The intent of the code_verifier / code_challenge is to meet the `PKCE` requirements. Rather than hiding that by calling it `symmetricKey` and therefore abandoning PKCE because the implementation is different or incomplete the better solution is to see if we can implement PKCE.

shamilovtim: Here is what we would need: https://datatracker.ietf.org/doc/html/rfc7636#section-4.1

shamilovtim: I have added a TODO comment on this topic. hopefully we can close this out. 

1. it was never intended to **only** be a symmetricKey
2. we should see if we can implement PKCE

.github/workflows/docs-ci.yml
LiranCohen: I think it's fair to bring up the additional burden and the fact that some comments are definitely unnecessary and only satisfying the bot (I'm guilty of this!)

But it seems out of scope for this PR, we should have that discussion and decide on better CI rules on their own merit.

packages/crypto/tests/utils.spec.ts
```diff
@@ -171,4 +172,58 @@ describe('Crypto Utils', () => {
...
+    });
+
+    it('generates an 8-digit PIN', () => {
+      const pin = randomPin({ length: 8 });
+      expect(pin).to.match(/^\d{8}$/);
+    });
+
+    it('generates an 9-digit PIN', () => {
+      const pin = randomPin({ length: 8 });
+      expect(pin).to.match(/^\d{8}$/);
```
LiranCohen, line 209:
```suggestion
      const pin = randomPin({ length: 9 });
      expect(pin).to.match(/^\d{9}$/);
```

packages/crypto/tests/utils.spec.ts
```diff
@@ -171,4 +172,58 @@ describe('Crypto Utils', () => {
...
+    });
+
+    it('generates an 10-digit PIN', () => {
+      const pin = randomPin({ length: 8 });
+      expect(pin).to.match(/^\d{8}$/);
```
LiranCohen, line 214:
```suggestion
      const pin = randomPin({ length: 10 });
      expect(pin).to.match(/^\d{10}$/);
```

.github/workflows/docs-ci.yml
shamilovtim: When we have that discussion i would love for one of us to go through and delete useless comments because they seriously take away from readability which is why I didn't want to ship them

packages/agent/src/oidc.ts
LiranCohen: I'm supportive of fixing this in a further integration PR.

If you could ADD this issue link to the TODO: https://github.com/TBD54566975/web5-js/issues/827
There will be a PermissionApi class that will help with the various grant operations.

packages/agent/src/oidc.ts
```diff
@@ -0,0 +1,703 @@
...
+    return message;
+  });
+
+  const permissionGrants = (await Promise.all(grantPromises)).flat();
+
+  return permissionGrants;
+}
+
+async function submitAuthResponse(
+  selectedDids: string[],
```
LiranCohen, line 632: Sanity: It seems like we are only ever using one DID, is there a reason for this to be an Array vs just `selectedDid: string`?

packages/agent/src/oidc.ts
```diff
@@ -0,0 +1,703 @@
...
+  authRequest: Web5ConnectAuthRequest,
+  randomPin: string,
+  dwn: AgentDwnApi
+) {
+  const delegateDid = await DidDht.create();
```
LiranCohen, line 637: The delegateDID should be a `did:jwk`, there's no need for it to have service endpoints or key-rolling.

packages/agent/src/connect.ts
```diff
@@ -0,0 +1,179 @@
...
+async function initClient({
+  connectServerUrl,
+  walletUri,
+  permissionRequests,
+  onWalletUriReady,
+  validatePin,
+}: WalletConnectOptions) {
+  // ephemeral client did for ECDH, signing, verification
+  // TODO: use separate keys for ECDH vs. sign/verify. could maybe use secp256k1.
+  const clientDid = await DidDht.create();
```
LiranCohen, line 19: Sanity: Does this need to be a `did:dht` or can we use a `did:jwk`?

LiranCohen (commented):
Getting super close! Just noticed a few small things.

packages/agent/src/oidc.ts
shamilovtim: Good call will fix

packages/agent/src/oidc.ts
shamilovtim: I agree and thats what I was doing at first (did:key or did:jwk) but frank recommended I use DHT for everything so now would be a rewrite to go back again. @thehenrytsai what do you think?

packages/agent/src/connect.ts
csuwildcat: It would be better if the DIDs returned to the client are DID JWKs

packages/agent/src/oidc.ts
csuwildcat: I think Frank may have meant the Agent DIDs, because those actually do need to sync. We should transition the permissioned DIDs given to apps to be DID JWK.

shamilovtim: he really did mean all dids but i might have taken what he was saying too literally and this is an exception

packages/agent/src/oidc.ts
shamilovtim: fixed

packages/agent/src/oidc.ts
shamilovtim: fixed

packages/agent/src/oidc.ts
thehenrytsai: Agree with @csuwildcat and @LiranCohen that the app only needs a DID that essentially holds a private key. This is the reason I questioned why this was originally an array of private keys instead of one in my original comment.

I am personally fine with any DID method as long as it does not require resolution in this use case. I care more about agreeing with the actual structure we are stamping as the standard for shipping the DID (with private key), and looks like `PortableDid` is the candidate of choice.

shamilovtim: ok so `did:jwk` with PortableDid?

have we abandoned `did:key` or whats the usecase for it anymore?

.github/workflows/docs-ci.yml
thehenrytsai: I feel the same pain and agree on some comments being pretty useless, and did not have time and passionate enough to sink on challenging the intent of the original owner/creator that sets up the doc rule and CICD.

The error/warning pollution in PRs was however distracting and annoying enough that I followed the existing intent and cleanup all existing warnings/errors, wasn't too painful as frankly Chad did most of the work.

I do agree with @LiranCohen that this is out of scope though happy to discuss about this in a separate issue PR. If you are passionate about this, feel free to create an issue and assign it to yourself to drive to resolution.

But before we reach consensus on changing the rules, we should follow them.

.github/workflows/tests-ci.yml
thehenrytsai: Fair point about hermit folder. Me dumb.

package.json
thehenrytsai: I have not had to run a `corepack` command, and not in favor of needing to run it to keep the newly added line up to date unless base on my understanding. Are you proposing that we use corepack instead of hermit? If not, suggesting to remove this.

Happy to jump in a call to be convinced otherwise.

shamilovtim: like I explained, hermit is using corepack. so this is functioning as intended. if we don't want corepack we can uninstall it.

shamilovtim: updated PR to remove corepack

packages/agent/src/connect-v2.ts
thehenrytsai: I'd agree to implementing PKCE (`code_challenge/verifier`) in a latter PR and would be down to take the work item too , however (sorry for beating the dead horse), why are we using the `code_challenge` as the symmetric encryption key? I thought the point of encrypting the request is so that the relay server (as well as unintended actor) is not able to decrypt the content? The `code_challenge` if I understand correctly, is meant to be visible to the relay server, thus using `code_challenge` as the encryption key would allow relay server to decrypt the content if/when PKCE is implemented. This is why I kept suggesting leaving `code_challenge/verifier` alone and just call the random bytes what it is: a symmetric key. Apologies if I am still not making sense, please educate me if I continue to misunderstand things.

.github/workflows/docs-ci.yml
shamilovtim: Made a ticket for it. Keep in mind it's not working for this PR so its checks will need to be bypassed with an admin merge (or set back to warning as I did in the PR)

packages/agent/src/connect-v2.ts
shamilovtim: ok that makes sense. so just leave the verifier unused?

.github/workflows/docs-ci.yml
thehenrytsai: Pretty sure it is working as intended. Happy to push a change to fix these after you are done with all the force-pushes so my fixes don't get overwritten.

package.json
thehenrytsai: Thanks for removing `corepack`. I have no idea why it was included in `hermit` and if we need it or not. My comment is purely about the line added in the PR, which seemed unnecessary because we never had it and things seems to worked "fine"....

packages/agent/src/connect-v2.ts
thehenrytsai: My suggestion is still calling it a key:

`const symmetricKey = utils.randomBytes(32);`

And call it a `key` or similar in wallet URI.

> (though I am curious if this pattern is something we invented in web5, or a known OIDC pattern.)

I am a noob to OIDC too, so I don't know if there are better "more compliant or established" ways to ship an encryption key hence my original comment above, but what I know is shipping/calling it as `code_challenge` is definitely not it.

packages/agent/src/oidc.ts
shamilovtim: fixed

packages/agent/src/connect.ts
shamilovtim: fixed

.github/workflows/docs-ci.yml
shamilovtim: once this is approved and then #830 is approved and merged into this I can revert this change

thehenrytsai: Word. Please do revert when #830 goes in. 🙏

packages/agent/src/connect.ts
```diff
@@ -32,12 +39,12 @@ async function initClient({
...
   // build the PAR request
   const request = await Oidc.createAuthRequest({
-    client_id             : clientDid.uri,
-    scope                 : 'web5', // TODO: clear with frank
-    code_challenge        : codeChallengeBase64Url,
-    code_challenge_method : 'S256',
-    permissionRequests    : permissionRequests,
-    redirect_uri          : callbackEndpoint,
+    client_id          : clientDid.uri,
+    scope              : 'web5', // TODO: clear with frank
```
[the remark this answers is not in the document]
thehenrytsai: Sanity: is it supposed to be `'openid did:dht'` or `'openid did:jwk'`?

thehenrytsai (dismissed):
🚀 🎖️ 🎊 🥳 🎆 🍾 

Optional consideration to merge #837 as I doubt #830 itself will get rid of tbdocs errors.

thehenrytsai (approved):

thehenrytsai (approved):
Third time lucky.

thehenrytsai (approved):
4th time is the charm.

csuwildcat (approved):

PR #743, opened 2024-07-11
Protocol definitions for `DwnDidStore`, `DwnKeyStore` and `DwnIdentityStore`

LiranCohen:
This PR requires `DwnDataStore` to use a `protocol` for the storage abstraction. The protocol is initialized when any data is `set`, the result of the initialization is cached in an in-memory map so that the DWN does not need to query for the installed protocol on each subsequent invocation of set.

This is split into two protocols, `identity-store` which is used for `DwnIdentityStore` and `DwnDidStore`
```typescript
 {
    protocol  : 'http://identity.foundation/protocols/web5/identity-store',
    published : false,
    types     : {
      portableDid: {
        schema      : 'https://identity.foundation/schemas/web5/portable-did',
        dataFormats : [
          'application/json'
        ]
      },
      identityMetadata: {
        schema      : 'https://identity.foundation/schemas/web5/identity-metadata',
        dataFormats : [
          'application/json'
        ]
      }
    },
    structure: {
      portableDid      : {},
      identityMetadata : {}
    }
  };
```

And `jwk-store` which stores the `privateJwk` for a given key
```typescript
 {
    protocol  : 'http://identity.foundation/protocols/web5/jwk-store',
    published : false,
    types     : {
      privateJwk: {
        schema      : 'https://identity.foundation/schemas/web5/private-jwk',
        dataFormats : [
          'application/json'
        ]
      },
    },
    structure: {
      privateJwk: {}
    }
  };
```

There can be improvements with these stores such as using tags instead of the cached index and scanning of all records, but that will be done in a separate PR after delegated/permissioned key functionality is complete.


packages/agent/src/store-data-protocols.ts
```diff
@@ -0,0 +1,40 @@
...
+      ]
+    }
+  },
+  structure: {
+    portableDid      : {},
+    identityMetadata : {}
+  }
+};
+
+export const JWKProtocolDefinition: ProtocolDefinition = {
```
thehenrytsai, line 26: Nit: 
```suggestion
export const JwkProtocolDefinition: ProtocolDefinition = {
```

thehenrytsai (approved):
Appreciate the PR description, making my first journey into the Store abstractions less treacherous. Code generally looks good, though can't say I understand everything.

I do find a lot of naming problematic, which is probably the source a lot of confusion. e.g.

- `DwnDataStore` is a complete reuse of term in `dwn-sdk-js` but mean different thing.
- the term "DID", "identify" and "tenant" all smell very similar, I wish there is one lineer documentation describing the intent/purpose of `DwnDidStore` and `DwnIdentityStore`.
- `getDataStoreTenant()` - at a glance seem so odd having to pass tenant to get tenant. LOL.

thehenrytsai (approved):

csuwildcat (approved):

PR #744, opened 2024-07-11
Delete Identity/DID/Key from DWN Stores

LiranCohen:
- `delete` method added to `DidApi` which deletes the `portableDid` stored in the `DidStore`
  - Optionally, if the `deleteKey` option is set to false the keys associated with the DID stored by the `KeyStore` are deleted. **default is true**
- `delete` method added to `IdentityApi` which deletes the `IdentityMetadata` stored in the `IdentityStore`
- `deleteKey` method added to the `LocalKeyManager` and `AgentKeyManager` interface.


csuwildcat (approved):

thehenrytsai (approved):
🚀

shamilovtim (approved):

csuwildcat (approved):

PR #745, opened 2024-07-12
make `rawMessage` public on the `Record` class

LiranCohen:
- `rawMessage` property on the `Record` class is not private
- skip `did:ion` test in the `credentials` package


shamilovtim (approved):

thehenrytsai (approved):
🐐

PR #750, opened 2024-07-15
DWN Endpoint during `Web5.connect()` DID Creation

LiranCohen:
Prior to this PR when a new DID is created using `Web5.connect()` 2 DWN endpoints would be chosen from a list of 6 that TBD hosts.

With this PR:
- `techPreview` option is removed in favor of `didCreateOptions` which accepts the same `dwnEndpoints` string array as an option
- If `dwnEndpoints` are not provided, the default `https://dwn.tbddev.org/latest` is assigned.


shamilovtim (dismissed):

packages/api/src/web5.ts
```diff
@@ -136,6 +141,12 @@ export type Web5ConnectOptions = {
    * See {@link TechPreviewOptions} for available options.
    */
   techPreview?: TechPreviewOptions;
```
thehenrytsai, line 143: Sanity: should this be removed also?

LiranCohen: @thehenrytsai, @csuwildcat wanted us to keep it there for now so that existing code doesn't break.

thehenrytsai (approved):
🐐 🚀

csuwildcat (approved):

PR #769, opened 2024-07-17
DWN Registration

LiranCohen:
Allows for DWN registration by passing in a `registration` object to `Web5.connect()`

The registration object takes `onSuccess()` and `onFailure()` methods, which are called respectively if registration succeeds or fails for the given DWN Endpoints that require registration.

Each are only called once if all succeed or any failures happen.

If none of the endpoints require registration neither are called.

### DIFF
--- a/packages/api/src/web5.ts
+++ b/packages/api/src/web5.ts
@@ -330,9 +330,6 @@ export class Web5 {
         // If a registration object is passed, we attempt to register the AgentDID and the ConnectedDID with the DWN endpoints provided
         const serviceEndpointNodes = techPreview?.dwnEndpoints ?? didCreateOptions?.dwnEndpoints;
 
-        // We only want to return the success callback if we successfully register with all DWN endpoints
-        // So we keep track of whether we attempted registration at all
-        let registrationAttempt = false;
         try {
           for (const dwnEndpoint of serviceEndpointNodes) {
             // check if endpoint needs registration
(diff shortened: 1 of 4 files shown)


thehenrytsai (approved):
Might want to fix the low hanging missing docs, else every PR going forward is going to be blocked (for people with no override power).

packages/api/src/web5.ts
```diff
@@ -326,30 +328,37 @@ export class Web5 {
...
+
             // register the agent DID
             await DwnRegistrar.registerTenant(dwnEndpoint, agent.agentDid.uri);
 
             // register the connected Identity DID
             await DwnRegistrar.registerTenant(dwnEndpoint, connectedDid);
+          }
 
-            // if no failure occurs, call the onSuccess callback
+          if (registrationAttempt) {
```
thehenrytsai, line 354: Is it possible the caller passes in a registration object, and no registration attempt is taken? According to the code path it looks it is possible, but in which case the caller would not be signaled at all regarding a outcome even though they explicitly said they wanted registration to take place, is this okay?

packages/api/src/web5.ts
```diff
@@ -326,30 +328,37 @@ export class Web5 {
...
       if (registration !== undefined) {
         // If a registration object is passed, we attempt to register the AgentDID and the ConnectedDID with the DWN endpoints provided
-
         const serviceEndpointNodes = techPreview?.dwnEndpoints ?? didCreateOptions?.dwnEndpoints;
-        for (const dwnEndpoint of serviceEndpointNodes) {
-          try {
 
+        // We only want to return the success callback if we successfully register with all DWN endpoints
+        // So we keep track of whether we attempted registration at all
+        let registrationAttempt = false;
```
thehenrytsai, line 335: Style nit: consider moving `registrationAttempt` into the try, because no one outside the try uses it.

Issue #680, opened 2024-06-06
[api] Unexpected error: Expected 1 but found 2 stored identities. When calling `Web5.connect()`


LiranCohen:
When using `Web5.connect()` without any options, there have been occurrences where the following error is thrown:
`Uncaught (in promise) Error: connect() failed due to unexpected state: Expected 1 but found 2 stored identities.`

When `connect()` is called and no identities are returned from the agent, a new identity is created.
There is some unknown scenario where `userAgent.identities.list()` is returning zero results, although there was one that existed in the past.

On a subsequent call of `connect()` it will then suddenly have 2 identities.
https://github.com/TBD54566975/web5-js/blob/main/packages/api/src/web5.ts#L205-L248

For the time being we will do a few things to remedy the stuck situation, as well as investigate further as to why this happens.
1. Check the status returned by `processRequest` within the `DwnDidStore.getAllRecords()` and throw an error along with the error detail if it does not return a 200 status. https://github.com/TBD54566975/web5-js/blob/main/packages/agent/src/store-did.ts#L48-L57
2. If the user passes a `connectedDid` and the list returns more than one identity, attempt to find the identity which matches the `connectedDid`, if it exists use that identity, otherwise fail.

csuwildcat:
Hopefully connectedDid is the DID identifier, not some more complex thing?

LiranCohen:
> Hopefully connectedDid is the DID identifier, not some more complex thing?

Yup! It is set to the `identity.did.uri` further down the file if there is no agent passed.

LiranCohen:
This was partially done during the connect flow.

Currently if multiple identities exist, the first in the index is returned. Closing in favor of better-defined behavior for multi-DID management.

PR #824, opened 2024-08-05
Adding `PermissionsApi` to agent and `permissions()` api to `Web5.dwn` with the ability to select delegate grants for `Web5.dwn`

LiranCohen:
This PR makes use of a mocked `Web5Connect` flow to process/store, select and use the grants when necessary.

I'm treating this as a first-pass, there will come more refactoring which will allow for connecting/exporting an existing Identity, selecting form multiple identities, requesting additional grants from a connected Did, etc.

If any of the provided `grants` during connect fail to process for whatever reason, all of the provided grants will be rolled back and the provided Identity will be cleaned up/deleted so that a subsequent Connect can be initiated.

Additionally this PR puts a bandaid on the `multiple identities` issue. It will simply default to the first identity in the array if multiple are provided instead of failing.
[Unexpected error: Expected 1 but found 2 stored identities. When calling Web5.connect()](https://github.com/TBD54566975/web5-js/issues/680)

A `syncManager` implementation will come after this PR, some of the current tests manually shuffle/install permissions and protocol installations between logical DWNs. When `syncManager` is updated a one-shot sync will take place after processing the grants in order to get up-to-date with the connected protocols.
[Implement Grants in SyncManger](https://github.com/TBD54566975/web5-js/issues/826)

~~This PR creates an experimental `grant` API in `@web5/api`, in a subsequent effort there will be a more well defined general purpose `PermissionsApi` that will likely live in the `agent`, and a better defined `grant` API within `@web5/api` that will allow creation of `requests`, `grants` and `revocations` as well as querying/selecting.
[Concise Grants APIs](https://github.com/TBD54566975/web5-js/issues/827)~~

When using the `dwn.records` API with grants, providing a `protocol` in the params is needed. In a subsequent PR this property will be required for all `records` usage as we are moving away from "flat space" records without a protocol assigned. This will have a lot of breaking tests and docs so it is best to do it in it's own PR to avoid bloat/urgency.
[Require protocol for dwn.records API #825](https://github.com/TBD54566975/web5-js/issues/825)

There should also be a separate effort to think about how both `protocolPath` scoped, as well as `context` scoped grants. More specifically the user would need to provide the `contextId` and/or `protocolPath` when issuing a `records.read()` or `records.delete()`, which don't normally have these properties. The `records.delete()` can be replaced with the `delete()` method now available on the logical `Record` object, which would help this, but `read` would still need to be provided a better way of selecting grants.
[Grants for a specific Context/Path](https://github.com/TBD54566975/web5-js/issues/828)

---

PermissionsApi
Introduces a PermissionsApi interface and an AgentPermissionsApi concrete implementation.

The interface implements the following methods fetchGrants, fetchRequests, isGrantRevoked, createGrant, createRequest, createRevocation as convenience methods for dealing with the built-in permission protocol records.

The AgentPermissionsApi implements an additional static method matchGrantFromArray which was moved from a PermissionsUtil class, which is used to find the appropriate grant to use when authoring a message.

dwn.connected
A Private API usedin a connected state to find and cache the correct grants to use for the request.

dwn.permissions
A Permissions API which implements request, grant, queryRequests, and queryGrants that a user can utilize

The Web5 permissions api introduces 3 helper classes to represent permissions:

PermissionRequest
Class to represent a permission request record. It implements convenience methods similar to the Record class where you can store(), import() or send() the underlying request record. Additionally a grant() method will create a PermissionGrant object.

PermissionGrant
Class to represent a grant record. It implements convenience methods similar to the Record class where you can store(), import() or send() the underlying grant record. Additionally a revoke() method will create a GrantRevocation object, and isRevoked() will check if the underlying grant has been revoked.

GrantRevocation
Class to represent a permission grant revocation record. It implements convenience methods similar to the Record class where you can store() or send() the underlying revocation record.


packages/agent/tests/dwn-permissions-util.spec.ts
```diff
@@ -106,19 +106,21 @@ describe('DwnPermissionsUtil', () => {
 
       const protocol = 'http://example.com/protocol';
 
-      await GrantsUtil.createRecordsGrants({
+      const deviceXRecordGrants = await GrantsUtil.createRecordsGrants({
```
LiranCohen, line 109: Remove the reliance on `fetchGrants` from this test, as it's out of scope for the util.

packages/api/src/web5.ts
```diff
@@ -174,6 +181,9 @@ export type Web5ConnectResult = {
   /** The DID that has been connected or created during the connection process. */
   did: string;
 
+  /** The DID that is used to sign messages on behalf of the connectedDID */
+  delegatedDid?: string;
```
shamilovtim, line 185: `delegateDid` (no d)

packages/api/src/web5.ts
```diff
@@ -196,6 +206,9 @@ export type Web5Params = {
 
   /** The DID of the tenant under which all DID, DWN, and VC requests are being performed. */
   connectedDid: string;
+
+  /** The DID that will be signing Web5 messages using grants from the connectedDid */
+  delegatedDid?: string;
```
shamilovtim, line 211: delegateDid (no d)

LiranCohen: 🙏 fixed in all places

csuwildcat (approved):

PR #830, opened 2024-08-09
cleanup crypto utils

shamilovtim:
web5 crypto had a package.json named export which causes pain for different bundlers, mocha, etc. removed it and replaced with a regular esm export. 

merge after #713


csuwildcat (approved):

LiranCohen (approved):

thehenrytsai (approved):
Remember to revert:

```
fail_on_warnings: true
fail_on_error: true
```

PR #838, opened 2024-08-13
cleanup crypto utils 

shamilovtim:
- fixes a bug in the codecov yml where the config was not being respected
- bump the codecov action version
- exports the crypto utils only one way through the `CryptoUtils` namespace (rather than 3 different ways)
- deletes a few unused methods

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -7,8 +7,11 @@ component_management:
       - type: project
         target: auto # auto compares coverage to the previous base commit
         threshold: 5% # allows a 5% drop from the previous base commit coverage
+        informational: true
       - type: patch
         target: 90
+        informational: true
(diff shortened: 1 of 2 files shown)


codecov.yml
```diff
@@ -7,8 +7,11 @@ component_management:
       - type: project
         target: auto # auto compares coverage to the previous base commit
         threshold: 5% # allows a 5% drop from the previous base commit coverage
+        informational: true
       - type: patch
         target: 90
+        informational: true
+        if_ci_failed: success
```
LiranCohen, line 14: Why is this being done again? Can this be removed to a different PR that's in scope?

LiranCohen (approved):

PR #846, opened 2024-08-14
bump dwnserver

shamilovtim:


LiranCohen (changes requested):
Should probably bump the `dwn-sdk-js` version in `agent` and in the dev dependencies of `api` also.

LiranCohen (approved):
🚀

PR #850, opened 2024-08-21
connect dwn integration

shamilovtim:


packages/api/src/web5.ts
```diff
@@ -158,6 +159,8 @@ export type Web5ConnectResult = {
    * {@link WalletConnectOptions} was provided.
    */
   delegateDid?: string;
+
+  delegateGrants: DelegateGrant[];
```
LiranCohen, line 163:
```suggestion
```

Let's remove this. We just added this as a temporary way to test and display on the requesting app something when we were pairing.

Now that the delegate sync part is in connect, we can demo it by writing a record as alice.

examples/wallet-connect.html
```diff
@@ -139,10 +139,15 @@ <h1>Success</h1>
...
+      {
+        interface: "Records",
+        method: "Read",
+        protocol: "http://profile-protocol.xyz",
+      },
     ];
 
     try {
-      const { delegateDid } = await Web5.connect({
+      const { delegateDid, delegateGrants } = await Web5.connect({
```
LiranCohen, line 150: delegateGrants won't exist as discussed in the connect comment. Also the references below.

```suggestion
      const { delegateDid } = await Web5.connect({
```

LiranCohen (commented):
Looks great! Just one small thing, we added that `delegateGrants` return from `Web5.connect()` as just our own visual testing. But now that delegate sync is in we can just write a record or something.

LiranCohen (approved):

PR #855, opened 2024-08-27
agent did api resolver cache

LiranCohen:
In some cases, specifically when more than one identity exists, the `connect()` method can take over a minute to load.

This is because DHT resolution takes places for each identity in the system when calling `identities.list()`.

I won't go into the details of why resolution happens, but it's part of getting the signer to the DWN tenant for each identity.

This PR implements @csuwildcat's cache implementation on the `AgentDidApi` class which takes into account which DIDs are managed by the agent (or the agent's own DID). These DIDs are never evicted from the cache, only updated versions are inserted after a successful resolution occurs.

Other DIDs are evicted as usual.

In addition to this I have increased the `recordId` reference storage cache to 21 days instead of 2 hours. These values don't change, so we should cache them for as long as possible. The max items is 1,000 and it will evict items based on last seen.  21 days was chosen because the maximum allowed timeout in Node is 24 days, I reduced it as a buffer.

This should speed up `connect()` time and all signing of messages throughout the system.


.github/workflows/alpha-npm.yml
```diff
@@ -21,15 +21,16 @@ jobs:
 
     env:
       # Packages not listed here will be excluded from publishing
-      PACKAGES: "agent api common credentials crypto crypto-aws-kms dids identity-agent proxy-agent user-agent"
+      # These are currently in a specific order due to dependency requirements
+      PACKAGES: "crypto crypto-aws-kms common dids credentials agent identity-agent proxy-agent user-agent api"
```
LiranCohen, line 25: At some point this workflow stopped being able to build. After the step `Generate alpha prerelease and bump package.json @web5/* versions` modifies the package versions, pnpm is running the builds in alphabetical order which is causing dependency issues (looking for a version that doesn't exist).

A temporary remedy is to build the packages sequentially given the order in which they depend on each other.

csuwildcat (approved):

PR #893, opened 2024-09-06
differentiate between Record creator and author

LiranCohen:
Add a property to the `Record` class that denotes the record's `creator`.

When the record is an initial write, the `creator` and `author` will return the same.

Created this issue to add query/filter for creator: https://github.com/TBD54566975/dwn-sdk-js/issues/800

### DIFF
--- a/.changeset/brown-experts-lay.md
+++ b/.changeset/brown-experts-lay.md
@@ -1,0 +1,5 @@
+---
+"@web5/api": patch
+---
+
+differentiate between Record creator and author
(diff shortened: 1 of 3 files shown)


csuwildcat (approved):

thehenrytsai (approved):

LiranCohen:
Add a property to the `Record` class that denotes the record's `creator`.

When the record is an initial write, the `creator` and `author` will return the same.

Created this issue to add query/filter for creator: https://github.com/TBD54566975/dwn-sdk-js/issues/800

PR #894, opened 2024-09-07
Protocol Query with or without grant, Configure with grant.

LiranCohen:
- protocol query with regular permission grant
  - if grant is not found, author query as delegate did to get any public protocols
- protocol configure with delegate grant
- `Permission` can now include `configure` which represents `ProtocolsConfigure` of a particular protocol
- `createPermissionRequestForProtocol` now includes a grant for `ProtocolsQuery` for the protocol.

### DIFF
--- a/packages/agent/src/oidc.ts
+++ b/packages/agent/src/oidc.ts
@@ -602,12 +602,16 @@ function encryptAuthResponse({
 }
 
 function shouldUseDelegatePermission(scope: DwnPermissionScope): boolean {
+  // Currently all record permissions are treated as delegated permissions
+  // In the future only methods that modify state will be delegated and the rest will be normal permissions
   if (isRecordPermissionScope(scope)) {
     return true;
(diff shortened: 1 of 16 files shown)


packages/agent/src/oidc.ts
```diff
@@ -600,6 +601,20 @@ function encryptAuthResponse({
...
+  if (isRecordPermissionScope(scope)) {
+    return true;
+  } else if (scope.interface === DwnInterfaceName.Protocols && scope.method === DwnMethodName.Configure) {
+    // ProtocolConfigure messages are also delegated, as they modify state
+    return true;
+  }
+
+  // All other permissions are not treated as delegated
+  return false;
+}
```
thehenrytsai, line 616: Add tests to cover these lines before merge?

thehenrytsai (approved):
Let's go.

csuwildcat (approved):

PR #897, opened 2024-09-09
replace Identity API `list` method with `lisetMetadata`

LiranCohen:
While working with restoring a a synced agent, I ran into an issue where the Identity API's `list` method requires both the messages from the agent's DID and the Identity itself, however that Identity may not have been registered to sync (yet or ever).

This doesn't solve the issue entirely, however I did notice that most times when listing the identities, you first want the metadata which gives you what you need to then `get` the `BearerIdentity`.

I've replaced `list` with `listMetadata` in this PR and updated the usage in `@web5/api`.

In a subsequent PR(s?) I will address the underlying issue associated with syncing an agent, or potentially refactoring the Identity API completely, however I wanted to put in this low-effort PR to unblock a path in a sane way.

### DIFF
--- a/.changeset/popular-monkeys-rush.md
+++ b/.changeset/popular-monkeys-rush.md
@@ -1,0 +1,8 @@
+---
+"@web5/identity-agent": minor
+"@web5/agent": minor
+"@web5/proxy-agent": minor
+"@web5/user-agent": minor
+---
+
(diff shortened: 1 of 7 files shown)


LiranCohen:
Decided on different approach. Closing.

PR #911, opened 2024-09-26
Refactor `IdentityApi` to be scoped to the agent as a tenant.

LiranCohen:
When thinking about importing/syncing identities across wallets (or agents) there is a lot of extra complexity associated with the different tenants who store the identities.

Within this PR:
- simplify `IdentityApi` to be agent-focused and storing both the DID and IdentityMetadata under the agent's tenant.
- Remove the need for `manage` in IdentityApi.
- DIDs that are created/imported are automatically pre-populated in the resolution cache.
- Update vulnerable packages

The DidApi remains unchanged and a default `did.create()` will store the DID document as the tenant of the newly created DID.

In a subsequent PRs I would like to take some of the logic from `AgentDidResolverCache` and move it to the DidStore. That way not only agent DIDs but any DIDs which are imported int the DID store will be respected as long-term DIDs and the DID Store itself will be refreshed with updated DID doc information from resolution. As well as introduce an `update` method to the DID / Identity APIs.


packages/agent/src/identity-api.ts
thehenrytsai, line 1: Request: can you add a little sentence above `class AgentIdentityApi` explaining the intent/purpose of this class for new comers? Thanks!!

thehenrytsai (approved):

csuwildcat (approved):

PR #913, opened 2024-09-26
Quick fix to tbdoc warnings that shows up in every PR

thehenrytsai:
Wrapping inside an exported class instead of exporting a composed object, benefits:

1. No change/break in downstream caller/user of the lib.
1. Avoids unnecessary documentation duplication.
1. Keeping the original intent of grouping the methods.


LiranCohen (approved):
🐐

PR #961, opened 2024-10-22
Fix issue where `dwn-store` records are not actually being updated

LiranCohen:


shamilovtim (dismissed):

packages/agent/src/store-data.ts
```diff
@@ -158,8 +158,8 @@ export class DwnDataStore<TStoreObject extends Record<string, any> = Jwk> implem
 
       // set the recordId in the messageParams to update the existing record
       // set the dateCreated to the existing dateCreated as this is an immutable property
-      messageParams.recordId = matchingRecordEntry.recordsWrite?.recordId;
-      messageParams.dateCreated = matchingRecordEntry.recordsWrite?.descriptor.dateCreated;
+      messageParams.recordId = matchingRecordEntry.recordsWrite!.recordId;
+      messageParams.dateCreated = matchingRecordEntry.recordsWrite!.descriptor.dateCreated;
```
shamilovtim, line 162: not important but you could hide the complexity in a type guard and not have to do the `!`

shamilovtim (approved):

PR #964, opened 2024-10-23
add prerequisites for expansion of connect flow

### LINKED ISSUE #748: wallet connect: use a separate key for ecdh
Right now wallet connect uses the same key for ecdh, signing, verification and impersonation. This should be split up in the future so that a unique key is used for ecdh.

shamilovtim:
### Context

The web5 connect flow previously relied on the delegateDid in order to encrypt, sign + verify and move the workflow from step to step.

While this worked for an MVP implementation to get the release out the door, it had a number of issues. The same DID should not have been used for sign/verify as is used for encryption and that same DID should also not be used for persistent delegation.

The code was implemented with TODOs knowing that this would need to be enhanced but the enhancements were slated as security enhancements only. 

When the "export" flow was proposed it turned out that this code would need to be fixed immediately. In the export flow the delegation is only transferred at the end of the connect flow. This means that you therefore cannot rely on the delegation for the creation of shared encryption keys. 

Therefore fixing this security issue became a priority and a prerequisite for the "export" flow.

### Implementation

1. Here I have implemented the suggested fixes by creating separate DIDs for signing and verification, encryption, and delegation. 
2. I have also mapped out the export flow, which will come in a separate PR.
3. I cleaned up some comments 
4. I deduped the connect types, which I was able to simplify significantly.
5. I moved the `permissions` helper into the `initClient` method to reduce the amount of boilerplate going on in the `Web5` class

### Breaking changes for wallet authors.

web5 connect's `getAuthRequest()` now returns an object which include both the authRequest and a DID:

```ts
{
  authRequest: Web5ConnectAuthRequest;
  clientEcdhDid: DidResolutionResult;
}
```

web5 connect's `submitAuthResponse()` now requires that the did received from `getAuthRequest()` is passed in to the method at position 4:

```ts
async function submitAuthResponse(
  selectedDid: string,
  authRequest: Web5ConnectAuthRequest,
  randomPin: string,
  clientEcdhDid: DidResolutionResult,
  agent: Web5Agent
) { ... }

### DIFF
--- a/packages/agent/src/connect.ts
+++ b/packages/agent/src/connect.ts
@@ -138,7 +138,6 @@ async function initClient({
     baseURL  : connectServerUrl,
     endpoint : 'callback',
   });
-  console.log('after callback endpoint');
 
   // build the PAR request
   const request = await Oidc.createAuthRequest({
@@ -168,7 +167,6 @@ async function initClient({
     kid : clientEcdhDid.document.verificationMethod![0].id,
     encryptionKey,
   });
-  console.log('after requestobjecjtwe');
 
   // Convert the encrypted Request Object to URLSearchParams for form encoding.
   const formEncodedRequest = new URLSearchParams({
@@ -187,7 +185,6 @@ async function initClient({
       'Content-Type': 'application/x-www-form-urlencoded',
     },
   });
-  console.log('after par');
 
   if (!parResponse.ok) {
     throw new Error(`${parResponse.status}: ${parResponse.statusText}`);
@@ -204,7 +201,6 @@ async function initClient({
     'encryption_key',
     Convert.uint8Array(encryptionKey).toBase64Url()
   );
-  console.log('after generatedwalleturi');
 
   // call user's callback so they can send the URI to the wallet as they see fit
   onWalletUriReady(generatedWalletUri.toString());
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: packages/agent/src/connect.ts [hunk 1]
```diff
@@ -40,33 +144,37 @@ async function initClient({
     baseURL  : connectServerUrl,
     endpoint : 'callback',
   });
+  console.log('after callback endpoint');
```
shamilovtim: didnt meant to commit these. will remove.

packages/agent/src/oidc.ts
```diff
@@ -171,12 +185,20 @@ export type Web5ConnectAuthResponse = {
  * 2. `authorize`: provider gets the {@link Web5ConnectAuthRequest} JWT that was stored by the PAR
  * 3. `callback`: provider sends {@link Web5ConnectAuthResponse} to this endpoint
  * 4. `token`: client gets {@link Web5ConnectAuthResponse} from this endpoint
+ * 5. `export`: (if `exported` is true) client will POST a {@link PortableDid}
+ * 6. `retrieve`: (if `exported` is true) wallet will GET the {@link PortableDid}
+ * 7. `export-token`: (if `exported` is true) wallet will POST the grants in order to finalize the flow.
+ * 8. `retrieve-token`: (if `exported` is true) client will GET the grants in order to finalize the flow.
```
shamilovtim, line 191: these are all renamed in the next PR to be more clear, so just ignore this part

